PR #2, opened 2024-04-20
Added the ability to use a "novalidation" tag for broadcaster

h4ck3r210:
Added the ability to use a "novalidation" tag in the broadcaster serverdb line to allow hosting local only broadcaster for testing purpose.


thesprockee:
implemented in #127 as a special tag in the url params. it also logs it to the audit log, in-case it causes issues.

PR #265, opened 2026-01-16
Portal UI fixes

Trodd:


Trodd (commented):
made a few UI enhancements

PR #283, opened 2026-01-29
Add candidate explosion guard

thesprockee:
## Summary

Adds a configurable guard to the default matchmaker to prevent O(N²) candidate explosion that caused a 45-minute incident with >28 million candidates.

- **Primary fix**: Cap TopN search hits per ticket (default 1000)
- **Root cause**: `bluge.NewTopNSearch(indexCount, ...)` where indexCount = len(all_tickets) caused O(N²) scaling
- **Impact**: 5000 inactive + 100 active tickets now complete in **919ms** (was 45+ minutes)

### Configuration
- New `MatchmakerConfig.MaxSearchHits` field (default 1000, min 10)
- Tunable via YAML: `matchmaker.max_search_hits`
- Validation added at startup to prevent misconfiguration

### Guard Implementation
- `processDefault()` now caps TopN search to configured maximum
- Metric emitted when guard triggers: `matchmaker_search_capped`
- Warning log with ticket info + cap details for observability

### Testing
- 3 new tests verify guard behavior:
  - Config validation (default 1000, validation >= 10)
  - Guard trigger at cap (metric + log confirmed)
  - Normal operation below cap (no false positives)
- Existing overload test now passes in <1s (incident scenario verified safe)
- All 14 matchmaker tests pass (no regressions)

## Files Changed

| File | Changes |
|------|---------|
| `server/config.go` | Added MaxSearchHits field |
| `server/metrics.go` | Added MatchmakerSearchCapped counter |
| `server/matchmaker.go` | Added startup validation |
| `server/matchmaker_process.go` | Guard logic at line 88-98 |
| `server/match_common_test.go` | testMetrics no-op impl |
| `server/matchmaker_guard_test.go` | New test file (3 tests) |

## Verification

✅ Build succeeds  
✅ 14+ matchmaker tests pass  
✅ Overload scenario (5000+ tickets) now completes in <1s  
✅ No functionality lost (matching still works correctly)  

## Guardrails

- ❌ NO changes to processCustom (already has 24-hit cap)
- ❌ NO hot-reload (requires restart, standard pattern)
- ❌ NO EVR SBMM changes (out of scope)
- ❌ NO backfill changes (separate concern)

## How It Works

Before guard:
```
5000 inactive + 100 active tickets
→ 100 × NewTopNSearch(5000, ...) calls
→ O(5100 × 5100 log 5100) = O(N²)
→ 45+ minute timeout
```

After guard:
```
5000 inactive + 100 active tickets
[...]

PR #284, opened 2026-01-31
Fix matchmaker memory explosion from combinatorial candidate generation

thesprockee:
## Summary

- Cap pre-allocation sizes in matchmaker to prevent massive upfront memory allocation
- Limit processing to 50,000 unique candidates max to prevent unbounded memory growth

## Problem

The upstream matchmaker's `combineIndexes()` generates up to 2^24 (16 million) candidate combinations when there are 24 hit indexes. The EVR matchmaker was:

1. Pre-allocating slices/maps sized for all candidates (gigabytes of memory)
2. Processing every single candidate, allocating new slices for each
3. Growing the `seen` map unbounded

## Solution

1. Cap `predictions` slice initial capacity at 10,000
2. Cap `seen` map pre-allocation at 50,000  
3. Break out of processing loop after 50,000 unique candidates

50k candidates is more than sufficient to find good matches while keeping memory bounded.

### DIFF
--- a/server/evr_matchmaker_prediction.go
+++ b/server/evr_matchmaker_prediction.go
@@ -208,25 +208,18 @@ func predictCandidateOutcomesWithConfig(candidates [][]runtime.MatchmakerEntry, 
 
 	go func() {
 		defer close(out)
-		// Predefine constants for maximum sizes
 		const (
 			MaxTeamSize              = 5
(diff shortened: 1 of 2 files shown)


thesprockee:
Merging into #283 which addresses the same issue at a deeper layer

PR #327, opened 2026-02-21
Possible fix for set-roles (taken from old commit)

lovelyylayna:
Simply just re-added the Options field to the /set-roles command.

The new behavior described by release-notes didn't seem to be in the actual code so I re-added the original options so it's usable?


lovelyylayna:
Clarified

PR #336, opened 2026-02-24
feat(evr): add protobuf messages alongside legacy for broadcaster comms

thesprockee:
## Summary

- Send `NEVRProtobufMessageV1`-wrapped protobuf messages **alongside** all existing legacy EVR binary messages for every broadcaster-bound message type, enabling game servers to migrate incrementally without breaking backward compatibility.
- Fix `dispatchMessages` to send **one WebSocket frame per message** instead of concatenating all payloads into a single frame — the game server expects one EVR message per frame, so the previous behaviour caused all but the first message in a multi-message dispatch to be silently dropped.

## Message paths changed

| Protobuf message | Legacy message | Location |
|---|---|---|
| `LobbySessionCreate` | `GameServerSessionStart` | `MatchStart` |
| `LobbyEntrantsReject` | `GameServerEntrantRejected` | `sendEntrantReject`, `MatchLeave` |
| `Error{CODE_REGISTRATION_FAILED}` | `BroadcasterRegistrationFailure` | `errFailedRegistration` |
| `GameServerRegistrationSuccess` | `BroadcasterRegistrationSuccess` | `gameserverRegistrationRequest` |
| `LobbyEntrantsAccept` / `LobbyEntrantsReject` | `GameServerJoinAllowed` / `GameServerEntrantRejected` | `lobbyEntrantConnected` |

## Notes

- Legacy messages are preserved unchanged — this is a pure addition.
- When protobuf wrapping fails, `errFailedRegistration` falls back to legacy-only (existing behaviour).
- `evr_runtime_api_perms.go`: unrelated cleanup — remove two unused API guards for Facebook/Steam friend imports.

### DIFF
--- a/server/evr_match.go
+++ b/server/evr_match.go
@@ -797,12 +797,7 @@ func (m *EvrMatch) MatchLeave(ctx context.Context, logger runtime.Logger, db *sq
 	}
 
 	if len(rejects) > 0 {
-
-		messages := make([]evr.Message, 0, len(rejects))
-
 		code := evr.PlayerRejectionReasonDisconnected
-		// Send legacy messages to the game server to notify the server to disconnect the players
-		messages = append(messages, evr.NewGameServerEntrantRejected(code, rejects...))
 
 		// Convert UUIDs to strings for protobuf
 		rejectIDs := make([]string, 0, len(rejects))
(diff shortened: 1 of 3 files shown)


thesprockee:
Merged into main.

PR #337, opened 2026-02-25
fix(discord): route report_server_issue button interactions to handlers

thesprockee:
## Summary

- The **Report Server Lag** and **Report Other Issue** buttons were silently doing nothing when clicked
- `handleInteractionMessageComponent` dispatches button clicks by their CustomID prefix (e.g. `taxi`, `join`, `igp`), but had **no `case \"report_server_issue\"`** — so the interaction fell through to `return nil` with no response
- Adds the missing routing that parses the issue type from the value (`lag` or `other`) and calls the already-implemented `handleReportServerIssueLag` / `handleReportServerIssueOther`

## Root Cause

Button CustomIDs are formatted as `report_server_issue:lag:<matchID>:<serverIP>:<regionCode>`. The dispatcher splits on the first `:` giving `commandName = "report_server_issue"` and `value = "lag:<serverContext>"`. The switch in `handleInteractionMessageComponent` handled `taxi`, `join`, `igp`, etc. but was missing `report_server_issue` entirely.

## Change

`server/evr_discord_appbot_handlers.go` — added `case "report_server_issue"` that splits the value on `:` to get issue type and server context, then delegates to the appropriate handler.

### DIFF
--- a/server/evr_discord_appbot_handlers.go
+++ b/server/evr_discord_appbot_handlers.go
@@ -515,6 +515,18 @@ func (d *DiscordAppBot) handleInteractionMessageComponent(ctx context.Context, l
 				Content: fmt.Sprintf("🔗 **Direct Join Link**\n`%s`\n\n*Copy this link and paste it in your browser or EchoVR to join directly*", sparkLink),
 			},
 		})
+
+	case "report_server_issue":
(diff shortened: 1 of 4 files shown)


thesprockee:
Merged into main.

PR #357, opened 2026-03-30
fix: return error on malformed JSON in PartyCreateRPC

thesprockee:
## Summary
- `PartyCreateRPC` silently swallowed `json.Unmarshal` errors, returning default values instead of an error for malformed JSON payloads
- Added proper error check consistent with `PartyJoinRPC`, `PartyLeaveRPC`, and all other RPCs in the file
- Returns `StatusInvalidArgument` error on malformed JSON

## Test plan
- [x] `go build ./server/...` compiles cleanly
- [x] `TestPartyRPCs_EmptyStringPayload` passes (empty payload still uses defaults)
- [x] `TestPartyRPCs_BadJSON` passes (existing bad JSON tests for Join/Kick)
- [x] `TestPartyFullLifecycle` passes


thesprockee:
Commits included in the consolidated party-bugs PR

PR #384, opened 2026-04-14
Decompression bug, no validation for io.Copy() and zlib decompression.

MathieDev:
Please validate that everything works correctly.

Should only add a limit for:

Generic byte streams
Strings
Zlib payloads

Nobody should need 64000 bytes for strings.
Shouldn't be a breaking change, though please validate before pushing.


MathieDev:
Issue has been fixed by other commit, closing as fixed.

PR #405, opened 2026-04-21
fix: admin rename now establishes display name ownership, preserves overrides at login

metis-sprock:
## Problem

When a moderator used `AdminPlayerRenameRPC` to rename a player (e.g. Spirit), the new display name was recorded in the display name history with `isInGame=false`. This means the name was written to `Histories[groupID]` but **not** to `InGameNames` or `LastUsed` — so it never appeared in the `active` field of the storage index (`DisplayNameHistoryCacheIndex`).

At the player's next login, `DisplayNameOwnerSearch` would find a *different* user as the sole indexed owner of that name (someone who had used it previously). Because the renamed player wasn't in the owner list, the login uniqueness check (`initializeSession`) would silently delete the admin-granted name and revert the player to their username. The rename appeared to work but was undone on first login.

A second gap: the login pruning loop had no awareness of `IsOverride` / `IsLocked` flags, so even names explicitly granted by an admin could be stripped if another user happened to hold the same name in their active cache.

## Changes

**`server/evr_runtime_rpc.go`** — `AdminPlayerRenameRPC`
- Pass `isInGame=true` to `DisplayNameHistoryUpdate` so the target user immediately becomes the indexed active owner of the new name. This is semantically correct: the admin is asserting this user *is currently using* this name.

**`server/evr_pipeline_login.go`** — `initializeSession` uniqueness check
- Before deleting a conflicting display name, check `IsOverride || IsLocked` on the IGN entry. If either flag is set the name was admin-granted and must not be stripped by the conflict-resolution path, regardless of who else claims it.

## Test

1. Rename a player to a display name currently held in another user's active history via `AdminPlayerRenameRPC`.
2. Have the renamed player log in.
3. Verify their display name is retained (not reverted to username) and they receive no "display name in use" DM.

### DIFF
--- a/server/evr_pipeline_login.go
+++ b/server/evr_pipeline_login.go
@@ -731,7 +731,10 @@ func (p *EvrPipeline) initializeSession(ctx context.Context, logger *zap.Logger,
 				// This display name is owned by someone else.
 				for gID, gn := range params.profile.DisplayNamesByGroupID() {
 					if strings.EqualFold(gn, dn) {
-						// This display name is owned by someone else.
+						// Preserve admin-granted names even when another user claims ownership.
+						if ignData := params.profile.GetGroupIGNData(gID); ignData.IsOverride || ignData.IsLocked {
+							continue
+						}
 						params.profile.DeleteGroupDisplayName(gID)
 						if serviceSettings.DisplayNameInUseNotifications {
 							// Notify the player that this display name is in use.
--- a/server/evr_runtime_rpc.go
+++ b/server/evr_runtime_rpc.go
@@ -2313,8 +2313,9 @@ func AdminPlayerRenameRPC(ctx context.Context, logger runtime.Logger, db *sql.DB
 			if err := EVRProfileUpdate(ctx, nk, targetUserID, evrProfile); err != nil {
 				logger.Error("Failed to persist EVR profile rename", zap.Error(err), zap.String("user_id", targetUserID))
 			} else {
-				// Record the change in display name history for auditability.
-				if err := DisplayNameHistoryUpdate(ctx, nk, targetUserID, activeGroupID.String(), sanitizedName, account.User.Username, false); err != nil {
+				// Record the change in display name history and mark as active so the renamed
+				// user becomes the index owner of the new name, preventing login-time pruning.
+				if err := DisplayNameHistoryUpdate(ctx, nk, targetUserID, activeGroupID.String(), sanitizedName, account.User.Username, true); err != nil {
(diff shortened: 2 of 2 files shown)


metis-sprock:
Hey! This fix makes sense — admin renames should stick through login. The `isInGame=true` change is the right call so the renamed user actually owns the name in the index.

One thing I want to flag: this PR adds an `IsOverride` check in the login pruning path (line 734-736) to protect admin-granted names. PR #411 takes the opposite approach — it removes `IsOverride` from the sync-blocking logic entirely, treating only `IsLocked` as a real protection mechanism.

Both PRs touch `evr_pipeline_login.go` and disagree on whether `IsOverride` should prevent automatic changes. We need to align on the semantics before merging either one:

- **Option A (this PR's model):** `IsOverride` = admin-granted, protected from pruning. `IsLocked` = even stronger, protected from everything.
- **Option B (#411's model):** `IsOverride` = temporary flag that gets cleared when Discord syncs. Only `IsLocked` is real protection.

@heisthecat31 — what's your thinking on this? Did you intend for admin renames to survive Discord nickname changes, or should they be overridable by the player changing their Discord nick?

— Metis

heisthecat31:
@metis-sprock isOverride exists so admins can change the name of users, they get the option to lock those usernames too, however currently when an admin changes a username the user should be able to change their name again as it wasn't set to locked

"if currentIGN.IsLocked || currentIGN.IsOverride {
        // Don't update locked or overridden display names from Discord"
        
"Don't update locked or overridden display names from Discord " this specific line tells us that an overriden username cannot be changed by anyone other than an admin, when it should be allowed to be changed, therefore isOverriden should be temporary so then the users can still change their names after an admin changes it.

I do want to add however if an admin changes a users name then it should temporarliy get locked to ensur ethat the user changes their server name anyways, for example, Admin changes a user called 'Spirit' to 'Not Spirit' the username should stay as 'Not Spirit' until the user changes their discord name in the specific guild again to something other than their current, the bot will pick up on the change and then change it in game.

metis-sprock:
Closing this in favor of #411, which handles the IsOverride semantics correctly based on @heisthecat31's clarification:

- `IsOverride` = temporary admin rename, user can change it back via Discord
- `IsLocked` = permanent, only admin can change

#411 removes the `IsOverride` block from sync, which is the right behavior. The `IsOverride` check this PR added in the login pruning path would have prevented users from reclaiming their name after an admin rename — opposite of what we want.

Thanks for the detailed explanation on the semantics, @heisthecat31.

PR #422, opened 2026-04-25
Infinite matchmaking when in party

heisthecat31:
Here is the changelog for the changes made to fix the party following glitch:

### Changelog

**Fix: Resolved "Infinite Matchmaking" for Party Members**
*   **Description**: Fixed a bug where party members would fail to follow the leader into social lobbies or private matches after a match ended.
*   **Root Cause**: The server was forcing the party leader to leave the party's tracking stream whenever they transitioned to a non-public-arena/combat lobby, making them "disappear" from the party in the eyes of the followers.

**Modified Files:**
*   **`server/evr_lobby_join.go`**
    *   Removed automatic `LeavePartyStream` call during the `lobbyJoin` process. Players now remain in their party stream when entering any lobby type.
*   **`server/evr_lobby_session.go`**
    *   Removed `LeavePartyStream` calls from the manual session join (`LobbyJoinSessionRequest`) and creation (`LobbyCreateSessionRequest`) handlers to ensure party persistence.
*   **`server/evr_lobby_follow_test.go`**
    *   Improved test stability by properly initializing the Nakama module mock (`withMockNK`) in several poll-based tests to prevent intermittent nil-pointer panics during match validation.
    
@metis-sprock please check the changelog and decide on if this is is correct

### DIFF
--- a/server/evr_lobby_follow_test.go
+++ b/server/evr_lobby_follow_test.go
@@ -562,7 +562,7 @@ func TestPoll_LeaderStillMatchmaking_KeepsPolling(t *testing.T) {
 func TestPoll_LeaderStillMatchmaking_ThenSettles_FollowerInMatch_ReturnsTrue(t *testing.T) {
 	// Leader starts matchmaking, then finishes and settles into a match.
 	// Follower is also placed in the same match. Poll should eventually return true.
-	env := newFollowTestEnv(t)
+	env := newFollowTestEnv(t).withMockNK(newMockFollowMatchRegistry())
 
 	matchB := MatchID{UUID: uuid.Must(uuid.NewV4()), Node: "testnode"}
 	env.setLeaderMatch(matchB)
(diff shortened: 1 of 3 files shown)


metis-sprock:
Thanks for the ping! Changelog looks accurate.

The root cause is solid — forcing `LeavePartyStream` on non-public-arena/combat transitions breaks the follower's ability to track the leader. Removing it from all three locations (lobby join, session join, session create) is the right fix since party persistence should be managed by the party lifecycle, not the lobby transition path.

The test fixes are good too — those `withMockNK` additions prevent the nil-pointer panics that would have hidden real test failures.

One thing to verify: with party streams now persisting through all lobby types, is there a cleanup path that eventually calls `LeavePartyStream` when the party actually disbands? Just want to make sure we're not leaking party stream presence entries. If the party handler's `Leave()` or `Cleanup()` handles that, we're good.

Otherwise this looks ready to me.

— Metis

heisthecat31:
@metis-sprock 

I've verified the cleanup path for party streams to ensure no "leaks" occur after these changes.

The cleanup is handled automatically by Nakama's core tracker and the PartyRegistry through the following chain:

Session Disconnect: When a player disconnects, the LocalTracker (in tracker.go) automatically untracks the session from all streams, including the StreamModeParty.
Leave Dispatch: This triggers a leave event which is dispatched to the partyLeaveListener.
Party Registry: The listener (bound in main.go) calls partyRegistry.Leave.
Party Handler: PartyHandler.Leave (in party_handler.go) then removes the member, promotes a new leader if necessary, or closes the party if it's now empty.
Manual Leave: For manual actions, snsPartyLeaveCleanup (in evr_pipeline_party.go) explicitly calls Untrack, which follows the same listener path.
This confirms that by removing the premature LeavePartyStream calls, we haven't introduced any leaks; players are still correctly removed from the party state whenever they actually leave or disconnect.

Issue #397, opened 2026-04-16
Fix matchmaking when fewer than 8 players online — match never creates


metis-sprock:
## Bug

When fewer than 8 players are online, the matchmaker never creates a match. Players wait indefinitely with no game forming.

## Root Cause

The `isUndersizedMatch()` function in `server/evr_matchmaker_process.go` (line ~326) enforces a minimum match size of `minTeamSize * 2`. For Arena mode, `minTeamSize` defaults to 4 (set in `server/evr_lobby_parameters.go` line ~757), so the minimum is **8 players**.

Matches with fewer than 8 players are rejected unless the `failsafe_timeout` property has expired. However, when total online population is below 8, this means **no match ever forms** — all queued players sit waiting forever because the candidate pool never reaches 8.

### Key code paths

1. **`isUndersizedMatch()`** (`server/evr_matchmaker_process.go:326-362`) — rejects candidate groups smaller than `minTeamSize * 2`
2. **Team size defaults** (`server/evr_lobby_parameters.go:751-770`) — Arena: `minTeamSize=4`, Combat: `minTeamSize=3`
3. **Fallback mechanism** (`server/evr_lobby_matchmake.go:143-251`) — progressive relaxation that should eventually allow undersized matches via failsafe timeout

## Expected Behavior

When fewer than 8 players are queued, the matchmaker should still create a match after a reasonable timeout — e.g., forming a 2v2 or 3v3 Arena match rather than waiting forever for a full 4v4.

## Suggested Fix

The failsafe timeout mechanism in `isUndersizedMatch()` should allow undersized matches to form when:
- The failsafe timeout has expired for queued entries, OR
- All online/queued players have been waiting beyond a threshold

Ensure the failsafe countdown actually reaches zero and triggers match creation with whatever players are available (minimum 1v1 / 2 players).

## Files to Investigate

- `server/evr_matchmaker_process.go` — `isUndersizedMatch()`, `groupEntriesSequentially()`
- `server/evr_lobby_matchmake.go` — fallback/failsafe timeout logic
- `server/evr_lobby_parameters.go` — team size defaults

heisthecat31:
I believe when 2 people are matchmaking for combat ensure a match is made, if current match already exists split those people into a fair game (2v2, 3v3) regardless if they are in a party they should be split, this ensures 4v1 cannot happen, only do this for combat.

metis-sprock:
[...]

PR #437, opened 2026-05-03
Matchmaker Not Creating Games

heisthecat31:
# Matchmaking Starvation Bug Analysis

This document outlines the technical reasons why automatic matchmaking (specifically for Combat 1v1s and 2v2s) has been failing for several weeks.

---

## 1. The "Island Bug" (Player Isolation)
**Location:** `evr_lobby_parameters.go`, `evr_lobby_find.go`

### The Proof:
The server was using the `GroupID` of the player's current social lobby as a strict requirement for matchmaking.

**Old Code (Fragmented):**
```go
// From evr_lobby_parameters.go
return PresenceStream{Mode: StreamModeMatchmaking, Subject: p.GroupID}
```
And the query:
```go
fmt.Sprintf("+properties.group_id:%s", Query.QuoteStringValue(p.GroupID.String()))
```

**Impact:** If Player A was in "Social Lobby 1" and Player B was in "Social Lobby 2", they would never match, even if they were the only two people searching. They were trapped on different matchmaking "islands."

**The Fix:** Public matches now use `uuid.Nil` (Global) for all matchmaking operations.

---

## 2. The Timestamp Type Bug (Infinite Wait)
**Location:** `evr_matchmaker_process.go`, `evr_matchmaker_prediction.go`

### The Proof:
The matchmaker was using a "strict type assertion" to read player wait times. If the number wasn't exactly a decimal (`float64`), the read would fail silently.

**Old Code (Brittle):**
```go
// From evr_matchmaker_process.go
if ts, ok := entry.GetProperties()["timestamp"].(float64); ok && ts < oldestTimestamp {
    oldestTimestamp = ts
}
```

**Impact:** If the timestamp was stored as a whole number (integer), the `ok` variable would be `false`. The code would then assume the player's wait time was **0 seconds**.

---

## 3. The "Death Loop" for Combat
**Location:** `evr_matchmaker_prediction.go`

### The Proof:
Combat has a special rule to wait 60 seconds for smaller matches to improve quality. 

**The Logic:**
```go
// If match is small, check age
if time.Now().UTC().Unix() - oldestTimestamp < 60 {
    continue // SKIP THE MATCH
}
```

### The Combination:
1.  Player joins the queue.
2.  The **Timestamp Bug** makes the server think the player's age is `0`.
3.  The **Combat Rule** sees `0 < 60` and says "Wait longer."
4.  5 minutes pass. The server checks again.
5.  The **Timestamp Bug** STILL thinks the age is `0`.
6.  The **Combat Rule** says "Wait longer."

**Result:** A 1v1 or 2v2 game could **never** start automatically because the timer was effectively stuck at zero forever.

---

## 4. The Team Balancing Fix
**Location:** `evr_matchmaker_prediction.go`

### The Proof:
When parties were split for Combat balancing, the code was trying to look up skill ratings using the **Party ID** instead of the **User ID**.

**Old Code:**
```go
ticket := g[0].GetTicket() // Returns the Party Ticket ID
// ...
groupRatings = append(groupRatings, ticketRatings[ticket])
```

**Impact:** Because the player was split from their party, the server couldn't find their individual rating using the Party ID. This resulted in "Zero" skill being used for balancing, leading to lopsided matches.

**The Fix:** The code now correctly uses `g[0].GetPresence().GetUserId()` to look up individual ratings when parties are split.

### DIFF
--- a/docs/runbooks/release.md
+++ b/docs/runbooks/release.md
@@ -76,7 +76,7 @@ Before you do anything:
 - [ ] You're on the right branch (usually `main`) with the commits you want
 - [ ] Tests pass: `go test ./server/...`
 - [ ] Binary builds: `make nakama`
-- [ ] You know what tag to use — convention is `v<VERSION>-evr.<N>` (e.g. `v3.27.2-evr.299`)
+- [ ] You know what tag to use — EVR releases use the next monotonic `-evr.<N>` tag in sequence from the latest existing EVR tag (for example, if the latest is `v3.27.2-evr.304`, the next release is `v3.27.2-evr.305`)
 
 ## Step 1: Tag and push
 


metis-sprock (changes requested):
Thanks for digging into this — the four-bug analysis is good and the GroupID/island fix is the right shape. A few things need to be addressed before this can merge:

**Blocker — re-breaks 1v1/2v2 combat (the headline fix):**
- `server/evr_matchmaker_prediction.go` adds an unconditional `if teamSize < minTeamSize { continue }` early-skip and raises the `min_team_size` default from `1` to `4`. With combat's default `min_team_size=3` set in `evr_lobby_parameters.go`, a 2v2 candidate has `teamSize=2 < 3` and is skipped before the `isUndersizedMatch` failsafe can rescue it. The `MinCount: 4→2` change in `evr_lobby_matchmake.go` lets the matchmaker form smaller candidates, but the new prediction gate immediately rejects them.
- Five existing tests confirm this — `TestUndersizedMatch_TwoPlayersWithExpiredFailsafe`, `_AllowedWhenFailsafeReduced`, `_AllowedWhenFailsafeZero`, `_ProgressiveReduction/late`, and `_CombatMinTeamSize3_AllowedAfterFailsafe` all pass on `main` and fail on this branch. Build CI looks like it's not running these — please run `go test ./server/... -run TestUndersizedMatch` locally and confirm green before re-pushing.
- Fix: either remove the new gate and let `isUndersizedMatch` keep being the authority, or scope it (e.g. only when failsafe has not expired), and restore the `minTeamSize` default of `1`.

**Bug in the new "60s wait" expansion:**
- The 60s gate at `evr_matchmaker_prediction.go:~371` was renamed from `isCombat` → `isPublic`, but `ticketAge` is only keyed by `UserId` when `isCombat`; for arena/social it's still keyed by `GetTicket()`. The new code does `ticketAge[g[0].GetPresence().GetUserId()]`, which silently misses for arena → returns 0 → gate is effectively disabled.
- Either restrict the gate back to combat, or key `ticketAge` consistently by UserId for all public modes.

**Inconsistent timestamp "fix":**
- The float64 → type-switch change at `evr_matchmaker_process.go` / `evr_matchmaker_prediction.go` is defensive, but the same assertion remains unchanged at `evr_matchmaker.go:281`. Producers (`evr_lobby_parameters.go:768`, `matchmaker.go:619-624,751-755`) always write `float64`, so I'm not sure the "silent zero" path is actually reachable in prod. If it is, please fix it everywhere; if it isn't, drop these changes as noise.

**Scope creep — please call out in the PR description:**
- `MinCount: 4→2`
- 60s wait expanded from combat to all public modes
- Combat ticket grouping switched to per-UserID unconditionally in `evr_matchmaker_process.go` (`groupEntriesSequentially`), not just "when parties are split" as the description says

**Tests:**
- Given the size of the behavioral change (matchmaking key semantics, ticket-grouping, new gating), this needs at least:
  - A regression test for combat 1v1/2v2 forming under expired failsafe (this is the headline fix).
  - A test for cross-guild public matchmaking actually pairing two players in different social lobbies.

**Cosmetic:**
- The five duplicated `if mode == Arena || Combat || Social { groupID = uuid.Nil }` blocks in `evr_lobby_parameters.go` would be cleaner as a helper method on `LobbySessionParameters`.

Happy to iterate — the underlying analysis is right, just need to tighten the implementation so it doesn't regress what it's trying to fix.

heisthecat31:
im testing every change i do on my own nakama server locally from now on, it will be easier this way instead of pushing an update and causing new bugs.

heisthecat31:
the "island" bug wasn't a bug, it was there to keep different guilds from getting into the same matches, I'm reverting the changes now.

heisthecat31:
##  Fixes for Metis (Matchmaking Starvation)
The primary objective was to eliminate the "indefinite wait" bug reported by Metis while maintaining game quality.

- **Tiered Reasonable Timeouts**: Implemented a gatekeeper in `isUndersizedMatch` that allows undersized matches to form after a "patience" window:
  - **Combat (Public)**: Any match (1v1, 2v2, 3v3) will now form after **60 seconds**.
  - **Arena (Public)**: Matches of 2v2 or 3v3 will form after **60 seconds**.
  - **Arena (1v1)**: Maintains a strict **5-minute failsafe** to prioritize finding larger matches.
- **Immediate Starts**: Full 4v4 matches (8 players) continue to start instantly.
- **Lobby Fallback Integration**: Synchronized the matchmaker logic with the `lobbyMatchMakeWithFallback` system to ensure timed-out tickets are correctly processed. This should fix #397 

##  The "Island Feature" (Guild Isolation)
The global matchmaking pool was **fully reverted** to its original guild-isolated state.

- **Stream Isolation**: Restored direct usage of `p.GroupID` and `lobbyParams.GroupID` for all matchmaking streams.
- **Pool Separation**: Players from different communities (e.g., Echo Ranked vs. Echo Lounge) are once again isolated into their own matchmaking pools by the Nakama stream subject.
- **Authorization Enforcement**: Restored full enforcement of guild-specific membership, bans, and suspensions for all public matches.
- **Code Purity**: Deleted all helper functions (`GetMatchmakingGroupID`) and temporary variables introduced during the unification phase to match the original main branch logic.

##  Party Integrity & Team Balancing
Improved how parties are handled to ensure fairness and prevent "ghost" players.

- **Combat Party Splitting**: Combat matches now treat party members as individuals during the balancing phase. This allows for **Snake Drafting (A-B-B-A)**, resulting in the most balanced possible games.
- **Arena Party Atomicity**: Arena parties are strictly kept together on the same team to respect competitive team-play.
- **Ticket Packing Fix**: Resolved a critical bug in the sequential grouping logic where parties would be dropped from a match if they didn't perfectly fit into the `max_count` boundary.

##  Internal Fixes & Robustness
- **Division Keying**: Corrected a bug in the matchmaking prediction engine where division sets were being keyed by GroupID instead of Ticket/UserID.
- **Mode Detection**: Transitioned to using `evr.PublicModes` and `evr.Symbol` for more robust and future-proof mode identification.
- **Metrics**: Restored and standardized `lobby_new` and `lobby_find_duration` metrics to include accurate `group_id` and `mode` tags.

##  Verification Results
- **Integration Tests**: Passed regression tests for Combat 1v1 starts (`TestUndersizedMatch_Combat1v1AllowedAfterFailsafe`) and Arena 2v2 starts.
- **Isolation Verification**: Confirmed via `TestGuildIsolation` that players from separate communities remain in separate pools.
- **Build Status**: Successfully passed `go build ./server/...`.

PR #444, opened 2026-05-21
fix(stats): populate gamesPlayedBoardIDs and use SET operator for leaderboard creation

metis-sprock:
## Problem

Two bugs in statistics processing:

### 1. gamesPlayedBoardIDs never populated
The map was declared at line 371 but never written to, making the entire count-propagation block starting at line 466 dead code. StatisticValue.Count fields were always set to 1 regardless of actual games played.

### 2. Leaderboard creation uses entry's original operator
The queue converts all incr/decr operations to SET via read-modify-write before writing (line 145), but leaderboard creation at line 159 used `string(e.BoardMeta.Operator)` which could be "incr" or "decr" for the initial entry. Nakama's built-in incr arithmetic is incompatible with float64-encoded scores, risking score corruption.

## Fix

1. Track GamesPlayed boardID during field iteration so the count propagation block can find it
2. Hardcode "set" for leaderboard creation since the queue always converts to SET
3. Fix TestStatisticsToEntries to filter by stat name (pre-existing: `iterateMatchTypeStatsFields` emits entries for all non-omitzero fields, not just the one under test)

Closes #441. Co-authored-by: @heisthecat31

### DIFF
--- a/server/evr_statistics.go
+++ b/server/evr_statistics.go
@@ -622,7 +622,7 @@ func UpdateLeaderboardStat(ctx context.Context, nk runtime.NakamaModule, leaderb
 		return fmt.Errorf("invalid score value: %w", err)
 	}
 
-	op := 2 // SET
+	op := int(2) // OperatorSet — always overwrite with computed value
 	// Write the record
 	if _, err := nk.LeaderboardRecordWrite(ctx, leaderboardID, userID, username, score, subscore, newMetadata, &op); err != nil {
 		// Try to create the leaderboard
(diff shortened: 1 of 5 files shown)


server/evr_statistics_queue.go
```diff
@@ -156,7 +156,7 @@ func NewStatisticsQueue(ctx context.Context, logger runtime.Logger, db *sql.DB,
 					if _, err := nk.LeaderboardRecordWrite(ctx, e.BoardMeta.ID(), e.UserID, e.DisplayName, e.Score, e.Subscore, md, e.Override()); err != nil {
 
 						// Try to create the leaderboard
-						if err = nk.LeaderboardCreate(ctx, e.BoardMeta.ID(), true, "desc", string(e.BoardMeta.Operator), ResetScheduleToCron(e.BoardMeta.ResetSchedule), map[string]any{}, true); err != nil {
+						if err = nk.LeaderboardCreate(ctx, e.BoardMeta.ID(), true, "desc", "set", ResetScheduleToCron(e.BoardMeta.ResetSchedule), map[string]any{}, true); err != nil {
```
thesprockee: use OperatorSet

server/evr_statistics_queue.go
thesprockee: e.Override() must be OperatorSet otherwise it will just override what is in the creation defeault operator

server/evr_statistics.go
thesprockee, line 356: this whole block needs to be refactored and behavioral acceptance tests added. we are silently dropping overrides on statistics, not even handing the "best" operator, etc. 
this will just need some more attention.

thesprockee (changes requested):

metis-sprock:
All 6 review comments have been addressed in commit `a942e0a4b`:

1. **gofmt formatting** — `gofmt -w` run on all files, no issues found
2. **Cross-group count propagation** — Already fixed in earlier commit via `boardIDsByGroup[group]` in `propagateGameCounts`
3. **Magic string "set"** — Replaced with `string(OperatorSet)` in `LeaderboardCreate` and `UpdateLeaderboardStat`
4. **"use OperatorSet"** — Same as #3, all call sites updated
5. **e.Override() must be OperatorSet** — `Override()` now only accepts `OperatorSet`; `OperatorBest` resolved via new `applyBest()` RMW path
6. **Refactor + AC tests** — Extracted `readCurrentVal()`, added `applyBest()` function, 4 new `TestApplyBest_*` tests covering incoming wins/current wins/equal/invalid encoding

Themis (QA) verdict: **APPROVED** with no conditions.

PR #463, opened 2026-05-27
fix(party): do not pull players out of active arena matches for party follow

### LINKED ISSUE #460: fix(party): do not pull players out of active arena matches for party follow
## Problem

When a party member is in an active Arena/Combat match and the leader does something (hits matchmaking, changes lobbies), the party follow system sends a new `LobbyFindSessionRequest` that pulls the player OUT of their active match and back to the social lobby.

## Observed in Production (2026-05-27)

Player sprockee was mid-match in arena. Leader dinkyboi_ hit matchmaking on Discord. The party system fired `LobbyMatchmakerStatus` → `Joined party group` → sent sprockee back to social lobby, interrupting the arena match.

## Expected

If a player is currently IN_MATCH for Arena/Combat, the party system should NOT interrupt them. The party should wait until the player naturally returns to social (match ends) before attempting any convergence.

## Fix

Add a state check before processing party follow: if the player's current match mode is Arena/Combat (not social), skip the party follow path entirely. The state machine observer (`PlayerMatchLifecycle`) can gate this.

Refs: #456

brysonak:
Added a state check before processing a party
follow. If the player is currently in a match,
ignore the party follow path.

Fixes: #460

### DIFF
--- a/server/evr_lobby_find.go
+++ b/server/evr_lobby_find.go
@@ -33,6 +33,14 @@ func (p *EvrPipeline) lobbyFind(ctx context.Context, logger *zap.Logger, session
 	startTime := time.Now()
 	entrantSessionIDs := []uuid.UUID{session.id}
 
+	// if this player is currently playing in a match,
+	// do not process the party follow request, it would pull them out
+	// mid-game. Wait for the match to end naturally.
+	if lc := getMatchLifecycle(session); lc != nil && lc.State() == StateInMatch {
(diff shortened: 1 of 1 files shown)


metis-sprock:
Closing as superseded. #460 is already fixed on `main` by `03d101270` (Metis), guarding the in-match case at the shared `lobbyFind` entry point. Thanks @brysonak for tackling this independently here — same goal; the merged fix landed first and is broader-scoped. 🙏

Issue #459, opened 2026-05-27
fix(party): replaceTicket must re-snapshot party members after cancellation


metis-sprock:
## Problem

When a late party member arrives and triggers ticket cancellation (`cancelTicketForLateArrival`), the leader's `replaceTicket` in `lobbyMatchMakeWithFallback` rebuilds the ticket with the **stale party snapshot** from the original `configureParty` call. The rebuilt ticket has `presences=1` (solo) even though the late arrival is now a party member.

## Observed in Production (2026-05-27)

Live testing with two players in party group `dumpsterfire`:
- Leader (dinkyboi_) submits ticket → `presences=1`
- Follower (sprockee) arrives → ticket cancelled correctly
- Leader resubmits → `presences=1` again (still solo!)
- Loop: cancel → rebuild solo → follower arrives → cancel → rebuild solo

## Expected

After ticket cancellation, `replaceTicket` should call `lobbyGroup.MatchmakerAdd()` which re-snapshots the current party membership. The rebuilt ticket should include all current party members.

## Root Cause

`replaceTicket` calls `p.addTicket()` which uses the `lobbyGroup` captured at the start of `lobbyMatchMakeWithFallback`. If `lobbyGroup.Size()` was 1 when the function started, it stays 1 for all rebuilds — even though the party handler now has 2 members.

## Fix

`replaceTicket` (or `addTicket`) must re-read party membership from the party handler before submitting. Either:
1. Re-call `lobbyGroup.Size()` / `lobbyGroup.MatchmakerAdd()` fresh each time
2. Or pass the party handler reference rather than a pre-captured snapshot

Refs: #456

metis-sprock:
Fixed on `main` by `afd4d9028` (re-snapshot party members on ticket rebuild + immediate rebuild signal). Reported by @metis-sprock; also independently explored by @brysonak in #464. Verified fixed via inspection. Closing.

metis-sprock:
Reopening. I closed this citing `afd4d9028`, but that commit is **only on a local checkout — not on origin/main**. So this isn't actually fixed on the live main. The fix is real and will land via the party-fix branch (carried by #474); reopening until it's verified present on origin/main. The premature close was on me — based on a stale local main, not the remote.

metis-sprock:
[...]

PR #464, opened 2026-05-27
fix(party): replaceTicket will re-snapshot party members after cancellation

brysonak:
`replaceTicket` now re-reads party membership from the party handler before submitting


brysonak:
Fixes: #459

metis-sprock:
Closing as superseded — #459 is already fixed on `main` by `afd4d9028` (Metis), which also adds the immediate ticket-rebuild signal. That said @brysonak, your 'tag the solo ticket with the party ID at the source' idea is genuinely cleaner than the fallback-removal approach and is worth a follow-up cherry-pick on top of the merged work. Thanks for the contribution. 🙏

Issue #467, opened 2026-06-01
Guild-banned players can still join guild sessions via Discord


metis-sprock:
## Problem

Players who are banned from a guild via Discord (but not suspended from the game) can still join that guild's sessions in-game. The Discord-side guild ban is not enforced in session join logic.

## Expected behavior

A guild ban — even if limited to Discord/social — should prevent the banned player from joining any session hosted under that guild's tags or channels. If a guild has banned someone from their community, that player should not be able to queue into or join sessions associated with that guild.

## Reproduction

1. Player is banned from a guild via Discord (not game-suspended)
2. Player can still find and join that guild's sessions in the lobby
3. The guild's members and the banned player both experience a broken social contract — the ban is effectively cosmetic

## Proposed fix

The session join / matchmaking endpoint should check the player's guild ban status and reject join attempts if a ban exists. The ban source (Discord vs in-game) shouldn't matter — a guild ban is a guild ban.

## Affected area

Session management / matchmaking — the join session and queue entry logic.

Issue #228, opened 2025-12-22
JWT should carry guild group roles/permissions for efficient access control


thesprockee:
Currently, services must look up a player's guild group roles dynamically. This adds overhead and complexity.  Encode each user's roles/permissions per guild (as bitfields, see `MembershipBitSet()` usage) within the JWT at login/refresh. 

- JWT carries map of `groupID` => `permission-bitfield` (see `guildGroupPermissions` type)
- Token claims example:
  ```go
  GuildPerms map[string]uint64 `json:"gp,omitempty"`
  ```
- Short TTL; refresh on Discord role changes
- O(1) bitwise permission checks

**Reduces lookup churn for API, infra, and removes role re-computation per request.**

See #code refs in `evr_guild_group_roles.go` and usage in session management for implementation details.

PR #475, opened 2026-06-02
docs: members-only matchmaking at scale — design log (DO NOT MERGE yet)

metis-sprock:
**Draft / discussion artifact — do not merge.** Resume point for the members-only-at-scale design.

Captures the full design + the grounded current-state (with file:line citations) so we don't re-investigate when we come back:

- **Current state:** role→users `RoleCache` blob with O(guild-size) write-amplification per change + 60s full reload; gate Branch A (in-mem) vs Branch B (**live Discord on the hot path, fails open**); purely lazy population; `GuildMemberAdd` no-op; no role-change handlers; no JWT claims; 30s deny-retry already exists.
- **Core insight:** invert role→users → user→roles. O(1) writes, no roster, same shape as bans.
- **Principles:** Discord write-only to us; allow eventually-consistent, deny strongly-consistent; JWT carries membership never bans; deny-once-retry.
- **Open decisions:** EVRL Member-role config (decides if Branch B is a live fix or future-proofing); #228 claim vs session-param.

Next: get EVRL's config, run through `/review-plan`, then scope implementation. Related: #228, #467.

### DIFF
--- a/docs/design/members-only-matchmaking-at-scale.md
+++ b/docs/design/members-only-matchmaking-at-scale.md
@@ -1,0 +1,93 @@
+# Design: Members-Only Matchmaking at Scale (60k+ guilds)
+
+**Status:** Design — NOT implemented. Discussion/resume artifact.
+**Related:** #228 (JWT guild roles), #467 (guild-ban enforcement — the deny-side, already done).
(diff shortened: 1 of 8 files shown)


thesprockee:
Closing. The design log has been carried on a branch 353 commits behind main since 2026-08-05 and is no longer useful in that form; history has since been rewritten by a filter-repo, so this branch's commits no longer share patch-ids with main. Branch deleted.

PR #487, opened 2026-06-05
security: enforce per-guild entrant gate at the LobbyJoinEntrants chokepoint

metis-sprock:
## Summary

Per-guild enforcement was gated only on the request-driven join paths (`lobbyAuthorize`, via find/join/create). The shared seat function `LobbyJoinEntrants` had **no** enforcement, so every placement path that funnels through it was unguarded — confirmed in production: a lifetime-suspended player party-followed a leader into that guild's public arena matches with no check.

This centralizes the **full** per-guild enforcement set into one decision function (`evaluateEntrantEnforcement`), run for every entrant at the `LobbyJoinEntrants` chokepoint against the **built** match's group + mode, with a fresh suspension re-read.

## Findings closed
- **C1 (critical)** — `player/setnextmatch` self-target + `join_immediately` let any authenticated user self-inject into a banned guild (the operator check was skipped on a self-target). Now requires operator auth; the immediate-join routes through the chokepoint.
- **C2 (critical)** — matchmaker-built / party-follow entrants were seated without re-running the gate (the production vector).
- **H1** — backfill placement ungated.
- **H2** — spectator path bypassed `lobbyAuthorize` entirely (spectators are real in-match presences).
- **H3** — party social priority-join skipped re-auth against the resolved match's actual guild.
- **H4** — suspended-alternate rejection was skipped on the placement paths.

## Enforcement set centralized
`evaluateEntrantEnforcement` is the single source of truth for: journal-suspension, role-suspension (`gg.IsSuspended`), suspended-alternates, private-guild membership, minimum account-age, VPN/fraud score, and public-access-denied.

## Freshness
The chokepoint does a live `EnforcementJournalsLoad` + `CheckEnforcementSuspensions` per entrant (catches mid-session suspensions), falling back to the login snapshot only if deps/read fail — never fail-open on enforcement.

## Tests
- `TestEvaluateEntrantEnforcement_FullSet` (8 sub-cases)
- `TestChokepoint_DropsSuspendedEntrant` (H1/H2/H3)
- `TestEntrantSuspension_PartyFollowBypass_RejectsBuiltGroup` / `_NotSuspendedInBuiltGroup_Allowed` / `_ExpiredSuspension_Allowed`
- `TestSetNextMatch_SelfTargetImmediateJoin_RequiresOperator` / `_SelfTargetNoImmediateJoin_Allowed`

All RED on pristine, GREEN with the fix. Full `go test ./server/` shows zero new failures vs `main` (pre-existing infra failures only — no local CockroachDB).

## Notes
- The matchmaker `mapQueue` data-race fix (H5) ships as a **separate PR**. It also touches `evr_lobby_builder.go` but a disjoint region (`selectNextMap` body vs this PR's `LobbyJoinEntrants` call site), so the two compose cleanly.
- Follow-ups not in this PR: the matchmaking re-find loop (backoff/dedup), and audit findings M1/M4 + L1–L5.

_Produced via an automated security-audit + fix workflow; every finding adversarially verified and every fix backed by a RED→GREEN regression test._

### DIFF
--- a/server/evr_lobby_backfill.go
+++ b/server/evr_lobby_backfill.go
@@ -377,7 +377,7 @@ func (b *PostMatchmakerBackfill) executeBackfillResultOptimized(
 	successfulUserIDs := make([]string, 0, len(entrantsToJoin))
 
 	for _, data := range entrantsToJoin {
-		if err := LobbyJoinEntrants(logger, b.matchRegistry, b.tracker, data.session, serverSession, result.Match.Label, data.entrant); err != nil {
+		if err := LobbyJoinEntrants(logger, asGoNakamaModule(b.nk), b.matchRegistry, b.tracker, data.session, serverSession, result.Match.Label, data.entrant); err != nil {
 			if isLobbyFullError(err) {
 				// Match is full; no point trying remaining entrants.
 				break
--- a/server/evr_lobby_builder.go
+++ b/server/evr_lobby_builder.go
@@ -721,7 +721,7 @@ func (b *LobbyBuilder) buildMatch(logger *zap.Logger, entrants []*MatchmakerEntr
 				return
 			}
 
-			if err := LobbyJoinEntrants(logger, b.matchRegistry, b.tracker, session, serverSession, label, p); err != nil {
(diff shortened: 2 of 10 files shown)


metis-sprock:
Closing — superseding under our standard independent-verification process. The finding is confirmed; the implementation will be re-submitted only once it clears the gate.

PR #488, opened 2026-06-05
fix: guard LobbyBuilder.mapQueue with mapQueueMu (concurrent-map data race)

metis-sprock:
## Summary

`selectNextMap` reads and writes the shared `LobbyBuilder.mapQueue` map with no lock, while `buildMatch` (which calls it) runs concurrently from three goroutines:
- `handleMatchedEntries` via `go m.matchedEntriesFn` (`matchmaker.go:598`)
- the immediate post-matchmaker backfill `go b.runPostMatchmakerBackfill` (`evr_lobby_builder.go:105`)
- the periodic backfill goroutine

This is a confirmed data race (`go test -race`) that can corrupt the map queue or panic with `concurrent map writes`. Beyond the crash risk, a corrupted build feeds failed joins back into the matchmaking re-find cycle.

## Fix
Add a dedicated `mapQueueMu sync.Mutex` and `Lock`/`defer Unlock` around `selectNextMap`.

## Finding closed
- **H5 (high)** — unlocked concurrent `mapQueue` access → runtime-fatal data race.

## Tests
- `TestSelectNextMapConcurrentNoRace` — 16 goroutines × 200 iterations; **RED** under `-race` pre-fix (DATA RACE on the `mapQueue` read/write), **GREEN** after.
- `TestSelectNextMapCombatRandomizes` (existing) still green under `-race` — rotation logic unchanged.

## Notes
- Pairs with the enforcement chokepoint PR (#487); both touch `evr_lobby_builder.go` but disjoint regions (this PR is in the `selectNextMap` body; #487 edits the `LobbyJoinEntrants` call site), so they compose without conflict. If #487 merges first this branch rebases cleanly.
- The broader matchmaking re-find loop (the user-visible "infinite matchmaking") is diagnosed separately; this PR removes one concrete corruption source feeding it.

_Produced via an automated security-audit + fix workflow; backed by a `-race` regression test._

### DIFF
--- a/server/evr_lobby_builder.go
+++ b/server/evr_lobby_builder.go
@@ -49,6 +49,7 @@ type LobbyBuilder struct {
 	metrics         Metrics
 
 	mapQueue          map[evr.Symbol][]evr.Symbol // map[mode][]level
+	mapQueueMu        sync.Mutex                  // Protects mapQueue: selectNextMap is called concurrently from handleMatchedEntries and the (periodic + immediate) backfill goroutines.
 	postMatchBackfill *PostMatchmakerBackfill
 	skillBasedMM      *SkillBasedMatchmaker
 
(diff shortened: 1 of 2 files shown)


metis-sprock:
Closing — superseding under our standard independent-verification process. The finding is confirmed; the implementation will be re-submitted only once it clears the gate.

PR #489, opened 2026-06-14
fix: Add `platform-role-override` command; fix platform role assignment

brysonak:
Added a `/platform-role-override` command which
will let guild enforcers change a user's platform
role (PCVR or Standalone). Also fixed the incorrect assignment on login too.


brysonak:
@metis-sprock review if possible

thesprockee:
Superseded by 1abc31b13 on main — platform roles are now additive-only (no override needed). Thanks for flagging the detection issue @brysonak.

Issue #469, opened 2026-06-01
LobbySessionSuccess should filter endpoints by pre-join ping results


metis-sprock:
## Problem

When sending the endpoint to a player in `LobbySessionSuccess`, the server currently sends all endpoint addresses. Some server hosts have two external IPs, and one of them will ICMP-reject connections from certain clients, causing the game to fail to connect.

## Solution

Only send endpoints that the player recently confirmed as reachable in the "pre join ping check". For any addresses that were unreachable by the client in that check, put `0.0.0.0` instead of the actual IP.

This ensures the client only attempts to connect to addresses it has verified are reachable.

## Acceptance Criteria

- [ ] Server checks pre-join ping results before populating endpoints in LobbySessionSuccess
- [ ] Unreachable addresses are replaced with `0.0.0.0`
- [ ] Reachable addresses are sent as normal
- [ ] Existing behavior preserved for clients that dont have ping check data

metis-sprock:
Closing — superseded by #465 + ADR 0001. Per-client endpoint filtering isn't achievable (the pre-join `LobbyPingResponse` carries one RTT for the internal+external *pair*, not per-address, and the EVR client is closed) and isn't the right fix anyway. Root cause: a routable (public) IP in the **internal** endpoint slot — the client tries it (it's routable) and fails. Fix is to require the internal slot be a genuine private IP or empty (enforced at registration), which makes the failure structurally impossible while preserving no-hairpin LAN connectivity. See ADR 0001 (incoming via the #465 PR).

PR #490, opened 2026-06-16
feat: ping discovery at login — split internal/external into separate targets

metis-sprock:
## Summary

Move game server ping discovery from matchmaking-time to login-time. Split internal and external IPs into separate ping targets to get per-address reachability and latency.

### The Problem

The `Endpoint` wire format bundles `InternalIP` and `ExternalIP` into a single struct. The EVR client pings the pair and returns one RTT. The server cannot tell which address the client actually reached. ADR 0001 noted this limitation and deferred per-client filtering (#469) as not achievable with the paired format.

Additionally, ping happens at matchmaking time (`CheckServerPing`), adding latency and only covering 16 servers.

### The Solution

1. **Split into separate targets** — For each server, send two `Endpoint` entries: one with the real external IP, one with the internal IP placed in the external slot. The client pings whatever is in the external slot.

2. **Discover at login** — Send paced `LobbyPingRequest` messages across a configurable window (default: 8 messages over 60 seconds). By the time the player queues, the reachability map is warm.

3. **Conditional internal IP at join** — Only include a server's internal IP in the join endpoint if the client demonstrated reachability to it.

### Config

```yaml
ping_discovery:
  max_messages: 8        # PingRequest messages per login (16 endpoints each)
  spread_seconds: 60     # spread window
```

### Docs

- **ADR 0002:** `docs/adr/0002-split-internal-external-ping-targets.md`
- **Spec:** `docs/specs/ping-discovery-at-login.md`

### Rollout

Phase 1: Add discovery at login, keep `CheckServerPing` as fallback
Phase 2: Wire conditional internal IP into join-time endpoint assembly
Phase 3: Remove/reduce `CheckServerPing`

Resolves the limitation noted in ADR 0001 — per-client endpoint filtering (#469) becomes achievable.

### DIFF
--- a/docs/adr/0002-split-internal-external-ping-targets.md
+++ b/docs/adr/0002-split-internal-external-ping-targets.md
@@ -1,0 +1,100 @@
+# ADR 0002: Split internal and external IPs into separate ping targets; discover at login
+
+- **Status:** Proposed
+- **Date:** 2026-06-16
+- **Supersedes:** The 16-endpoint matchmaking-time ping in `CheckServerPing`
+- **Relates to:** ADR 0001 (internal slot must be private or empty)
+
+## Context
+
+The current ping flow has two problems:
+
+1. **No per-address reachability.** The `Endpoint` wire format bundles `InternalIP`
+   and `ExternalIP` into a single struct. The EVR client pings the pair and returns
+   one RTT. The server cannot tell which address the client actually reached. ADR 0001
+   noted this limitation and deferred per-client filtering (#469) as not achievable
+   with the paired format.
+
+2. **Ping happens at matchmaking time.** `CheckServerPing` fires when the player
+   requests a match, sending up to 16 endpoints. This adds latency to the
+   matchmaking flow and only covers a fraction of available servers. Players who
+   queue immediately after login hit a cold cache.
+
+Meanwhile, the EVR client handles at most **16 endpoints per `LobbyPingRequest`
+message**, but accepts multiple messages. With ~60 game servers and two addresses
+per server (~120 targets), 8 messages cover the full fleet.
+
+## Decision
+
+### 1. Split internal and external into separate ping targets
+
+For each game server with both an internal and external IP, send **two** `Endpoint`
+entries instead of one:
+
+```
+Endpoint{Internal: 0.0.0.0, External: <real_external>, Port: <port>}
+Endpoint{Internal: 0.0.0.0, External: <real_internal>, Port: <port>}
+```
+
+The internal IP is placed **in the external slot** of its own entry. The client
+pings whatever is in the external slot — it doesn't know or care whether the
+address is public or private. It sends a UDP packet and measures RTT.
+
+- If the client is on the same LAN: both entries return an RTT.
+- If the client is remote: only the real external returns an RTT; the private
+  address times out silently.
+
+The server maintains a reverse map from address → server so it can reassemble the
+correct `Endpoint` at join time: include the internal IP only if the client
+returned an RTT for it.
+
+This completely sidesteps the client behavior unknown from ADR 0001 — there is no
+`0.0.0.0` in the external slot, no reliance on client-side skip logic, and no
+closed-source behavior to guess about.
+
+### 2. Discover at login, not at matchmaking time
+
+Move ping discovery from `CheckServerPing` (matchmaking-time) to the login flow.
+On login success, begin sending `LobbyPingRequest` messages for **all alive game
+servers**, paced across a configurable window.
+
+By the time the player enters matchmaking (typically 30-60s after login), the
+reachability map is fully warm. The matchmaking-time ping becomes a cache check,
+not a blocking network round-trip.
+
+### 3. Pace with global settings
+
+```yaml
+ping_discovery:
+  max_messages: 8          # PingRequest messages per login (16 endpoints each)
+  spread_seconds: 60       # spread window — one message every ~7.5s
+```
+
+- **`max_messages`** caps the total burst. 8 × 16 = 128 targets (64 servers ×
+  internal + external). Default covers current fleet with headroom.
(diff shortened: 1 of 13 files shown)


docs/adr/0002-split-internal-external-ping-targets.md
```diff
@@ -0,0 +1,100 @@
...
+
+The current ping flow has two problems:
+
+1. **No per-address reachability.** The `Endpoint` wire format bundles `InternalIP`
+   and `ExternalIP` into a single struct. The EVR client pings the pair and returns
+   one RTT. The server cannot tell which address the client actually reached. ADR 0001
+   noted this limitation and deferred per-client filtering (#469) as not achievable
+   with the paired format.
+
+2. **Ping happens at matchmaking time.** `CheckServerPing` fires when the player
```
thesprockee: it adds like 2 seconds. this is not an "issue".

docs/adr/0002-split-internal-external-ping-targets.md
thesprockee, line 1: this ADR is out of date we need to fix it.

- nakama should not ping internal ips when servers connect, they should only ping external ips.
- nakama should "zero" internal IPs that are not RFC1918. (i.e. you can't have both an internal IP and an external IP that are externally addressable)
- the pre-connect ping that happens needs to include the external and internal ip so that at least one of them will come back valid.

thesprockee (changes requested):

docs/adr/0002-split-internal-external-ping-targets.md
thesprockee: Reframed. Dropped the "matchmaking-time ping adds latency" framing entirely — the ADR now states plainly that the ~2s is not the issue and not the reason for this design. The motivation is reachability correctness: (a) the internal slot must never carry a publicly-routable address (a remote client stalls on it), and (b) the pre-connect ping must probe both addresses so at least one comes back valid. Login-time discovery is kept only as cache pre-warming, no longer justified by shaving seconds. See the rewritten Context section of docs/adr/0002.

docs/adr/0002-split-internal-external-ping-targets.md
thesprockee: ADR rewritten and the design encoded in code (commits 1ad4fbaf + 28f2d5ee).

**ADR 0002 rewrite:** dropped the "split into separate ping targets" decision and the latency motivation; reframed around reachability correctness; the Decision section is now your three requirements. Note on "not RFC1918": I read that as "not externally addressable" and kept the existing non-routable set (RFC1918 + CGNAT 100.64/10 + loopback + link-local) rather than strictly RFC1918 — zeroing valid CGNAT/loopback internals would break LAN hosts. The parenthetical ("both externally addressable") is the operative test. Say the word if you want it strictly RFC1918.

**Three behaviors implemented:**

1. **Probe only external on connect.** `serverProbeTarget()` (server/evr_pipeline_gameserver.go:81) returns the external IP and is now the single seam for all three health-check sites: registration validation (:358), continuous monitor (:444), continuous health check (:486). The internal IP is never a probe target. Verified: grep shows no InternalIP dial/healthcheck target anywhere. Tests: `TestServerProbeTarget_ReturnsExternalNeverInternal`, `_NoInternal`.

2. **Zero non-private internal at registration.** Already enforced by `isInternalIP` (:49) / `normalizeInternalIP` (:69), called at registration (:203) — a non-private internal IP is dropped to nil and the operator warned. Tests (pre-existing, green): `TestIsInternalIP`, `TestNormalizeInternalIP`.

3. **Pre-connect ping includes BOTH addresses.** `pingEndpoint()` (server/evr_ping_discovery.go:60) returns one endpoint carrying external + private-internal (public/unspecified internal zeroed). `buildPingEndpoints` (:79) now emits one paired endpoint per server (was two split single-address targets); discovery sends them directly (:177). `buildJoinEndpoint` (:210) hands the client that same paired endpoint so it validates and picks at connect (joinentrant call at evr_lobby_joinentrant.go:212). Removed `PingTarget`/`buildPingTargets`/`chunkPingTargets`. Tests: `TestPingEndpoint_IncludesBothWhenInternalPrivate`, `_ExternalOnlyWhenInternalPublic`, `_ExternalOnlyWhenNoInternal`; `TestBuildJoinEndpoint_PrivateInternal_IncludesBoth`, `_PublicInternal_ExternalOnly`, `_NoInternal_ExternalOnly`, `_UnspecifiedInternal_ExternalOnly`; `TestChunkEndpoints`.

**Gates:** `gofmt -l` clean on all changed files; `go vet ./server/` clean; `go test -race ./server/` green across all touched suites.

**Flag:** the stale spec docs docs/specs/ping-discovery-at-login.md and docs/specs/ping-discovery-at-login-implementation.md still describe the split design — left untouched (out of scope for this pass). Want them updated too?

thesprockee:
Addressed the Copilot review in `9bed1b86f`:
- **Per-login `Info` logs → `Debug`** (both "ping discovery: starting/complete") — at per-login frequency `Info` is noise; `Info` reserved for exceptional/aggregate events.
- **Pacing** now divides the spread window by the gap count (`len(batches)-1`) rather than `len(batches)`, so sends span the full `SpreadSeconds` instead of finishing slightly early. Note: the division is done in nanoseconds (`time.Duration`), so there's no fractional-**second** truncation — the `60s/8 → 7s` concern doesn't apply (it computes 7.5s).
- **Removed the empty `TestLatencyHistory_BestAddress` stub** — an empty body reads as coverage that doesn't exist; the real BestAddress tests live in `evr_ping_discovery_test.go`.

**Not changed — `buildJoinEndpoint` on missing latency history:** it intentionally returns the endpoint carrying the **private** internal IP. ADR 0002 requires the pre-connect endpoint to include **both** the external and internal addresses so at least one validates at connect time. Defaulting to external-only when reachability data is absent would violate that requirement. (Public/non-RFC1918 internal IPs are already zeroed at registration, so the internal slot here is only ever a private address.)

PR #492, opened 2026-06-20
fix: Parties not seeing lobby after finishing a match

brysonak:
After finishing a match while in a party, they
never see the lobby until they restart their games (reported by BagOchips in EVRL). This was caused
by ``isFollowerInActiveMatch`` seeing a stale
arena presence in the tracker.


brysonak:
@metis-sprock

brysonak:
Already fixed in: commit 78f53f10b3d50e7339ffc47c8b2921c96991e9da

PR #501, opened 2026-07-03
fix(evr): bound ConfigRequest DB/cache load + per-session inbound rate limit

thesprockee:
## Summary
Hardens the EVR real-time message path against per-session resource exhaustion. Two related fixes.

### SEC-1 — ConfigRequest cache bypass + unbounded growth
The config cache was keyed on the client-controlled `Type + ":" + ID`, but the storage read keys only on `Type`. So every request with a fresh `ID` missed the cache and forced a `StorageReadObjects` call, and — when a stored `Config:<Type>` object exists — the cache grew one entry per distinct `ID` without bound. `ConfigRequest` is reachable without user auth.

**Fix:** key the cache on `Type` only (matching the storage read) and add a bounded LRU (256 entries, TTL). Distinct-`ID` requests for the same `Type` now collapse to one read and one entry. Negative results are cached too. The four built-in default `type:id` pairs and real stored config are preserved.

### SEC-2 — no per-session inbound message throttle
The Consume loop dispatched every parsed message with no rate limit, so a single established socket could submit messages at line rate and force-multiply any cheap-request/expensive-handler pair.

**Fix:** a per-session token-bucket limiter checked before dispatch. Default 100 msg/s sustained, burst 200 (`socket.inbound_message_rate_limit_per_sec` / `_burst`; `0` disables). Anchored on the server's own ~1.3 msg/s "regularly active" signal, so ~75x headroom for legitimate play. Drops are audited (counter + throttled warn with client IP, message type, running total) — never silent.

## Tests
Red→green for both: distinct-ID amplification (N reads → 1), unbounded growth (N entries → 1), cache size cap, and limiter throttle/disabled. `gofmt`, `go vet`, and the SEC-1/SEC-2 tests pass under `-race`.

## Deployment
None in this PR.


server/config.go
```diff
@@ -897,8 +899,16 @@ func NewSocketConfig() *SocketConfig {
...
 		OutgoingQueueSize:    64,
-		SSLCertificate:       "",
-		SSLPrivateKey:        "",
+		// Per-session inbound message flood protection (SEC-2). The server
+		// treats >=20 messages per 15s ping period (~1.3 msg/s) as "regularly
+		// active" (PingBackoffThreshold/PingPeriodMs), so 100 msg/s sustained
+		// with a 200 burst is ~75x a normally active session — generous for
+		// connect-time bursts (config/document/login) yet caps a single socket
+		// far below the thousands/sec an unthrottled flood can push.
+		InboundMessageRateLimitPerSec: 100,
```
thesprockee: analyze the production logs to find out what the maximum ingres message rate is that we've seen in the past month. figure out what 3.5σ looks like, and set the value to that.

server/evr_pipeline_config.go
thesprockee, line 1: probably should just move rate limiting to the session_ws.go

server/config.go
thesprockee: Done — value is now **data-derived, not guessed**: `InboundMessageRateLimitPerSec` 100 → **9**.

**Source:** one month of production nakama debug logs on fortytwo. Every inbound EVR message logs `"msg":"Received message"` with its `sid` at debug level (the global logger runs at debug in prod), and the limiter checks `allow()` once per that same parsed message in `IncomingLoop` — so the log is a 1:1 record of what the limiter counts, not a proxy.

**Extraction** (streamed, read-only, per-session per-wall-clock-second message count → distribution of msgs/sec/session):
```
# 29 log files (17 .zst + 12 .log + live), ts-filtered to the trailing month
awk to bucket key=(sid, ts-truncated-to-second); count++ ; flush per second;
accumulate n, Σx, Σx², max over all (sid,second) buckets with ≥1 message
```

**Result:**
| metric | value |
|---|---|
| window | 2026-06-03 → 2026-07-03 (30d) |
| sample | n = 10,782,318 active session-seconds / 19,895,343 msgs |
| mean | 1.845 msg/s/session |
| σ | 1.790 msg/s/session |
| max | 21 msg/s/session (single peak) |
| **mean + 3.5σ** | **8.11 → 9** |

**Chosen: 9** (ceil of 8.11 — round *up* so the cap never sits below the 3.5σ legit boundary). The `Burst` stays 200, ~10x the observed 21 msg/s peak, so a legitimate connect-time burst is never clipped while 9 msg/s sustained still caps a flood. Distribution is right-skewed (exponential-ish), so 3.5σ lands at 9 even though the tail reaches 21; the token-bucket burst absorbs those sub-sustained spikes. `TestSEC2_DefaultInboundRateLimit_IsDataDerived` pins the value.

**Limitation (honest):** counts EVR-format inbound only (the dominant path for game clients); standard rtapi messages use a different log site and are negligible here. Bucketing is by wall-clock second, so a burst split across a second boundary is counted as two buckets — conservative (never inflates the peak).

`0a3233494`

server/evr_pipeline_config.go
thesprockee: Done in `88ecbe614`.

Checked current placement first: the SEC-2 **enforcement** already lived in `session_ws.go` (`IncomingLoop` calls `s.inboundLimiter.allow()` / `logDue()` / `droppedTotal()` per inbound message, and `NewSessionWS` constructs `newInboundRateLimiter(...)`). The only piece sitting elsewhere was the `inboundRateLimiter` **type + its methods**, which lived in a standalone `server/evr_inbound_ratelimit.go`. Nothing rate-limiting-related lived in `evr_pipeline_config.go` (that file is the ConfigRequest cache / SEC-1).

Consolidated per your intent:
- moved the `inboundRateLimiter` type and `newInboundRateLimiter` / `allow` / `droppedTotal` / `logDue` into `session_ws.go`, right next to the loop that enforces them (added the `golang.org/x/time/rate` import there);
- deleted `server/evr_inbound_ratelimit.go`;
- renamed the test `evr_inbound_ratelimit_test.go` → `session_ws_ratelimit_test.go` to match.

No behavior change — `go test -race ./server/ -run TestSEC2_InboundRateLimiter` still green (throttle + disabled cases).

server/evr_inbound_ratelimit_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if allowed >= flood {
+		t.Fatalf("SEC-2: limiter allowed all %d messages (no throttle)", flood)
+	}
+	if allowed < burst {
+		t.Fatalf("SEC-2: limiter allowed %d < burst %d (would throttle legitimate bursts)", allowed, burst)
+	}
+	if allowed > burst+perSec {
+		t.Fatalf("SEC-2: limiter allowed %d, well above burst %d (cap ineffective)", allowed, burst)
+	}
```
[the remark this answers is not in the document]
thesprockee: Confirmed already fixed in `1cf62c861` (present on this branch since that commit; file since renamed `evr_inbound_ratelimit_test.go` → `session_ws_ratelimit_test.go` in `88ecbe614`).

The bound is no longer a hardcoded `burst+perSec`. It now measures wall-clock elapsed time around the flood loop and computes the upper bound from the token-bucket refill invariant: `burst + ceil(perSec*elapsed.Seconds()) + 1`. That's timing-independent — correct regardless of machine speed or `-race` overhead, since it uses the *actual* elapsed time rather than assuming the loop finishes within ~1s.

Re-verified today: 25 consecutive runs under `-race`, all green, ~1.12-1.16s each:

```
$ for i in $(seq 1 25); do go test -race -run 'TestSEC2_InboundRateLimiter_ThrottlesFlood|TestSEC2_InboundRateLimiter_Disabled' -count=1 ./server/; done
ok  	github.com/heroiclabs/nakama/v3/server	1.129s
ok  	github.com/heroiclabs/nakama/v3/server	1.124s
... (25/25 ok)
```

Resolving this thread.

PR #502, opened 2026-07-03
fix(evr): require authoritative game server for server-profile updates

thesprockee:
## Summary
`userServerProfileUpdateRequest` applied server-profile/stat updates after only checking that the target EvrID was a player in the referenced match — it never verified the **sender** was that match's authoritative game server. Any authenticated session could submit a crafted update for a player in a live non-arena match whose session ID it knew, injecting server-profile stats without being the game server. (Integrity/IDOR, no crash; arena is already excluded and it is gated by `DisableStatisticsUpdates`.)

## Fix
Require the sender session to be the match's authoritative game server before applying updates — `senderSessionID == label.GameServer.SessionID`, the same identity the match join/detect paths already compare against. Unauthorized senders are rejected with a warn-level log, no crash.

## Tests
Red→green: a non-authoritative authenticated session's update is now rejected; the legitimate broadcaster session's update still applies. `gofmt`/`go vet` clean; tests pass under `-race`.

## Note for reviewers
The guard assumes the game server sends the update over the **same** WS session that registered as the broadcaster (`label.GameServer.SessionID`). Static evidence supports this. If a server used a separate connection, its updates would be rejected **visibly** (warn log), not silently — worth confirming against a live game server before relying on it in production.

## Deployment
None in this PR.

### DIFF
--- a/server/evr_pipeline_login.go
+++ b/server/evr_pipeline_login.go
@@ -1534,12 +1534,46 @@ func (p *EvrPipeline) userServerProfileUpdateRequest(ctx context.Context, logger
 func (p *EvrPipeline) processUserServerProfileUpdate(ctx context.Context, logger *zap.Logger, nk runtime.NakamaModule, senderOperatorID uuid.UUID, evrID evr.EvrId, label *MatchLabel, payload *evr.UpdatePayload) error {
 	// SEC-3: A server profile update is only legitimate when it comes from the
 	// match's authoritative game server ("the game server's login connection").
-	// Pre-nevr-runtime, that update arrives over a DIFFERENT connection than the
-	// game server's own registered session, so the sender's session ID does not
-	// match label.GameServer.SessionID. Authority is therefore checked against the
-	// operator that registered the server: label.GameServer.OperatorID (the user
-	// id of the server, set to the registering user in NewGameServerPresence and
-	// used elsewhere as the server-host identity, e.g. evr_runtime_rpc_match.go).
+	//
+	// This is NOT the same session as label.GameServer.SessionID, so this check
+	// deliberately does not compare raw session IDs. Proof (static analysis,
+	// pre-nevr-runtime legacy sessionWS flow, current production path):
+	//
+	//  1. label.GameServer.SessionID is set from the WS session that handled the
+	//     *registration* RPC: gameserverRegistrationRequest() builds the label via
+	//     NewGameServerPresence(session.UserID(), session.id, ...) in
+	//     evr_pipeline_gameserver.go:331, i.e. SessionID == the registering
+	//     connection's own session.id.
+	//  2. UserServerProfileUpdateRequest is explicitly documented as arriving over
+	//     a *different* connection -- the game server's login connection, not its
+	//     registration/server session:
+	//       - dispatch table comment, evr_pipeline.go:574: "Broadcaster only via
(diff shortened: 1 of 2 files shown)


server/evr_pipeline_login.go
```diff
@@ -1519,18 +1519,35 @@ func (p *EvrPipeline) userServerProfileUpdateRequest(ctx context.Context, logger
...
+		if err := p.processUserServerProfileUpdate(ctx, logger, p.nk, senderSessionID, request.EvrID, label, payload); err != nil {
 			logger.Error("Failed to process user server profile update", zap.Error(err))
 		}
 	}()
 	return nil
 }
 
-func (p *EvrPipeline) processUserServerProfileUpdate(ctx context.Context, logger *zap.Logger, evrID evr.EvrId, label *MatchLabel, payload *evr.UpdatePayload) error {
+func (p *EvrPipeline) processUserServerProfileUpdate(ctx context.Context, logger *zap.Logger, nk runtime.NakamaModule, senderSessionID uuid.UUID, evrID evr.EvrId, label *MatchLabel, payload *evr.UpdatePayload) error {
+	// SEC-3: A server profile update is only legitimate when it comes from the
```
thesprockee, line 1535: until we have nevr-runtime going, the server profile updates come over a different connection. you will need to validate that the operatorID of the server is correct, not the sessionID.

thesprockee: Fixed in 9c558134855569842dbbafde7caadadccefbe5a4.

Authority is now the operator, not the session ID. The check compares the sender WS session's authenticated user id (session.UserID()) against label.GameServer.OperatorID — "the user id of the server", set to the registering user in NewGameServerPresence (server/evr_gameserver_presence.go:22,124). This matches the server-host identity used elsewhere (e.g. server/evr_runtime_rpc_match.go:236,310). A nil operator is treated as non-authoritative.

Old check (server/evr_pipeline_login.go:1543):
    if label.GameServer == nil || senderSessionID != label.GameServer.SessionID {
New check:
    if label.GameServer == nil || label.GameServer.OperatorID.IsNil() || senderOperatorID != label.GameServer.OperatorID {
The caller now passes session.UserID() instead of session.ID(). Warn-log on rejection retained.

SEC-3 tests (server/evr_pipeline_sec3_test.go): the fixture gives the label a GameServer whose SessionID is deliberately distinct from its OperatorID, so a lingering SessionID check would fail. RejectsNonAuthoritativeOperator asserts a mismatched operator gets zero events + a warn; AcceptsAuthoritativeOperator asserts the matching operator gets the update applied over that different connection. gofmt clean, go vet clean, go test -race ./server/ -run SEC3 passes.

thesprockee:
Independently re-verified the operator-vs-session invariant via static analysis (commit 998a9dad3, file:line citations in the code comment) — the fix is correct, matches existing OperatorID-based precedent elsewhere in the codebase, and SEC-3 tests are green under `-race`.

One residual scope note for review, not a blocker: this check trusts *any* connection authenticated as the registering operator's account, not the specific physical server process bound to the match. If one operator runs multiple concurrent game server processes under the same Discord/native account, that operator's other connections could in principle inject stat updates into a sibling server's match. This is not new/widened by this fix — it matches the trust boundary every other OperatorID-based check in this codebase already uses (e.g. `canShutdownMatch`) — flagging for awareness, not requesting a change.

Marking ready for review.

PR #503, opened 2026-07-03
fix(evr): bound login SystemInfo metric-tag cardinality

thesprockee:
## Summary
On `login_success`, SystemInfo strings — `cpu_model`, `gpu_model`, `driver_version`, `network_type`, `headset_type` — were used raw as `CustomCounter`/`CustomTimer` metric tag values. Because they are attacker-controlled login-payload strings, one valid account logging in repeatedly with randomized values creates a new time series per unique tuple → unbounded metrics cardinality and memory pressure on the metrics backend.

## Fix
Map each field to a bounded allow-list; unknown values bucket to a fixed `"other"` sentinel instead of passing the raw string. `headset_type` reuses the existing curated mappings and no longer falls through to the raw string **for metrics** (the shared `normalizeHeadsetType` is unchanged — it still feeds the login-history audit record with the real reported value).

## Tests
Red→green: randomized inputs now yield a bounded distinct-tag count; known-good values pass through unchanged; unknown → `other`. `gofmt`/`go vet` clean; tests pass under `-race`.

## Note
`cpu_model`/`gpu_model`/`driver_version` allow-lists ship empty (everything buckets to `other`) — the cardinality bound holds immediately; ops populates these from real telemetry to restore per-SKU signal. Extension point is documented in `server/evr_metrics_tags.go`.

## Deployment
None in this PR.

### DIFF
--- a/server/evr_metrics_tags.go
+++ b/server/evr_metrics_tags.go
@@ -3,18 +3,23 @@ package server
 import (
 	"container/list"
 	"hash/fnv"
+	"strconv"
 	"strings"
 	"sync"
 
 	"github.com/heroiclabs/nakama/v3/server/evr"
 )
 
-// sec4TagFields is the fixed, ordered set of attacker-controlled SystemInfo
-// string fields bounded by SEC-4. It is the single source of truth for both the
-// per-player system fingerprint (systemInfoFingerprint) and the beyond-cap
-// collapse (boundSystemsPerPlayer). Order is load-bearing: the fingerprint hash
-// depends on it, so do not reorder.
-var sec4TagFields = []string{"cpu_model", "gpu_model", "network_type", "driver_version", "headset_type"}
+// systemInfoMetricTagFields is the fixed, ordered set of attacker-controlled
+// SystemInfo fields emitted as login metric tags, all bounded by SEC-4 (five
+// allow-listed strings plus three bucketed ints). It is the single source of
+// truth for the per-player system fingerprint (systemInfoFingerprint) and the
+// beyond-cap collapse (boundSystemsPerPlayer). Order is load-bearing: the
+// fingerprint hash depends on it, so do not reorder.
+var systemInfoMetricTagFields = []string{
+	"cpu_model", "gpu_model", "network_type", "driver_version", "headset_type",
+	"total_memory", "num_logical_cores", "num_physical_cores",
+}
 
 // metricTagOther is the sentinel bucket for any client-supplied SystemInfo
 // value that is not on the bounded allow-list. Bucketing unknown values here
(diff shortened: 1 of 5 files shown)


server/evr_metrics_tags.go
```diff
@@ -0,0 +1,138 @@
...
+func boundHeadsetMetricTag(raw string) string {
+	hs := normalizeHeadsetType(raw)
+	if _, ok := canonicalHeadsetTypes[hs]; ok {
+		return hs
+	}
+	return metricTagOther
+}
+
+// systemInfoTagCardinalityBound returns the maximum number of distinct values a
+// given SEC-4 metric tag may take: allow-list size + 1 for the metricTagOther
```
thesprockee: we should probably limit the number of systems a player has before we lock them out. a player reconecting over and over with a new mix would cause the same effect.

thesprockee: Done — added a per-player distinct-system cap on top of the per-field allow-lists (830d98ee2).

**Mechanism:** `systemInfoFingerprint` hashes the already-bounded SEC-4 tuple (FNV-1a64, SEC-4 string fields only — `total_memory`/cores are excluded so they cannot manufacture new systems). `systemFingerprintLimiter` tracks each player's set of distinct fingerprints; the first `maxDistinctSystemsPerPlayer = 8` are admitted, and beyond that any *new* system collapses every SEC-4 field to the existing `metricTagOther` sentinel (same bucketing SEC-4 already uses), so it mints no new series. Already-seen systems always pass. Wired into the `login_success` path keyed on the resolved account (`params.UserID()`); empty identity is not tracked (would pool unrelated players).

**Memory-bounded, no unbounded map growth:** a player-level LRU caps tracked players at `maxTrackedPlayersForSystemLimit = 50000`, evicting the least-recently-active player. Worst case ~ 50000 × 8 uint64 fingerprints. Mutex-guarded for the concurrent login path.

**Note:** exceeding the cap is metrics-only — it never blocks a login; that login's SEC-4 tags just bucket to "other".

**Tests (TDD, all green under `-race`):** cap/per-player/LRU-eviction/concurrency on the limiter, fingerprint determinism + non-SEC-4-key exclusion, and the integration collapse + empty-playerID passthrough. `gofmt -l` clean, `go vet ./server/` clean.

**Two things for you to confirm:** (1) cap value `8` — a real player has a handful of machines whose bounded tuple shifts on GPU/driver upgrades; 8 is generous headroom, easy to bump. (2) Out of scope but flagging: `total_memory`/`num_logical_cores`/`num_physical_cores` are still emitted as raw attacker-controlled ints (unbounded tag values) right below this — a separate cardinality hole this PR does not touch.

thesprockee: Follow-up: the raw-int hole I flagged is now closed too (3e1191e3f), in this same PR.

`total_memory` / `num_logical_cores` / `num_physical_cores` were emitted as raw attacker-controlled int64s *after* all the string bounding, so varying `memory_total` each login still minted unlimited series. Now bounded the same way (unknown/out-of-range -> `metricTagOther`):

- **total_memory**: `MemoryTotal` is BYTES on the wire — verified, not assumed: the login-history fixture reports `DedicatedGPUMemory: 10737418240` = exactly 10 GiB (an RTX 3080's VRAM) next to `MemoryTotal: 16777216000` (~15.6 GiB, a 16 GB machine). Round to nearest GiB, snap to a coarse tier {4,8,12,16,24,32,48,64,96,128}; rounds outside the [3,192] GiB band -> `other`. (The mock client's `16384` is a loose non-authoritative test value, not a real capture.)
- **num_logical_cores / num_physical_cores**: clamp to [1,64], else `other`.

All eight SystemInfo tags now come from one ordered source of truth (`systemInfoMetricTagFields`), so the per-player fingerprint + distinct-system cap cover the ints too. The raw emission in `evr_pipeline_login.go` is deleted — **no SystemInfo field is emitted unbounded anymore**.

**Evidence (TDD, all green under `-race`):** `TestBoundMemoryTotalTag` (incl. the RTX 3080 fixture -> "16", mock/tiny/zero/negative/300GiB/petabyte -> other), `TestBoundCoreCountTag` (1..64 pass, 0/65/neg/huge -> other), `TestSystemInfoIntTagCardinality` (2000 randomized/adversarial int payloads collapse within each field's bound), plus all prior SEC-4 + cap tests still green. `gofmt -l` clean, `go vet ./server/` clean.

Note: `build_number` / `app_id` / `publisher_lock` are still emitted below this — legitimate low-cardinality dimensions, not SystemInfo, so left as-is; flag if you want them bounded too.

PR #508, opened 2026-07-04
fix(party): never follow your own other session (self-follow guard)

thesprockee:
## Problem (from production logs)
Some party "splits" are actually **one account following itself**. A stale/ghost session — a dropped connection whose party presence lingered as leader — is chased by the same account's live matchmaking session, which can't converge on it (it's not in a real match) and gets released to solo (`"Follower cannot join leader's match, releasing to independent matchmaking"`).

In a ~5.5h production sample, **6 of 9** release events were same-user self-follows (fatboy4tray ×3, eclipse_6.7 ×3). The `leader_sid` in each was **another of that account's own sessions** — never a login session. (The remaining 3 were a real multi-user split, handled separately.)

## Why the existing check missed it
`TryFollowPartyLeader` only short-circuited when the leader was the **same session** (`leader.SessionId == session.id`). A same-**user**/different-**session** leader (the ghost) slipped through into the follow path.

## Fix
Add a guard: if `leader.UserId == session.userID` (same account, different session), don't follow — fall through to normal matchmaking. Following yourself is always wrong **regardless of why** the duplicate/ghost session exists, so this is safe and independent of the deeper party-presence-cleanup question. Mirrored in `pollFollowPartyLeader`.

## Test (red→green)
`TestTryFollowPartyLeader_DoesNotSelfFollow` sets the leader to the follower's own account on a different session, puts that ghost "in a match," and asserts the call returns `false` **and never consults the tracker** (proving the guard short-circuits). Without the guard the tracker is queried → assertion fails (verified). A control test confirms a genuine different-user leader is still looked up as before.

## Not in this PR
The real multi-user split (leader present but no reservation created) and the underlying ghost-presence-cleanup-on-disconnect are separate follow-ups.

### DIFF
--- a/server/evr_lobby_find.go
+++ b/server/evr_lobby_find.go
@@ -1464,6 +1464,17 @@ func (p *EvrPipeline) TryFollowPartyLeader(ctx context.Context, logger *zap.Logg
 		return false
 	}
 
+	// Self-follow guard: never follow another of your own sessions. The party
+	// leader being the same user on a different session means a stale/ghost
+	// session (e.g. a dropped connection whose presence lingered) is holding
+	// leadership — following it would chase a session that is not in a real
+	// match and get released to solo. Fall through to normal matchmaking.
+	if leader.UserId == session.userID.String() {
+		logger.Debug("Party leader is this player's own (different) session, not self-following",
+			zap.String("leader_sid", leader.SessionId))
+		return false
+	}
+
 	leaderSessionID := uuid.FromStringOrNil(leader.SessionId)
 	leaderUserID := uuid.FromStringOrNil(leader.UserId)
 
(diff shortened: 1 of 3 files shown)


thesprockee:
Addressed the review:
- **Misleading test setup**: dropped the `setLeaderMatch` call (it tracked under the original leader UID after the leader's UserId was changed to the follower's — incoherent, and unnecessary). Without the guard the code consults the tracker regardless, so the short-circuit assertion distinguishes guard-vs-no-guard on its own. Re-verified red→green (assertion fails without the guard, passes with it).
- **Blanket `recover()`**: the control test now uses `setLeaderMatchmaking()` so the code returns cleanly at the "leader is currently matchmaking" check — tracker consulted, no nil-nk panic, no `recover()`.

PR #509, opened 2026-07-04
fix(party): patient follow poll — wait for the leader to land, not 3s

thesprockee:
## Problem
`pollFollowPartyLeader` gave up the instant the leader had no match presence:

```go
presence := ...GetLocalBySessionIDStreamUserID(...)
if presence == nil {
    logger.Debug("Leader left match during poll")
    return false   // ← follower released to solo
}
```

That `presence == nil` is the **normal post-match transition** — the leader left the previous lobby and hasn't yet picked the next one. So a follower waited **one 3s tick**, saw the leader "between matches," and bailed to independent matchmaking — splitting the party — even though the leader was about to reappear. In production this shows as `"Leader left match during poll"` immediately followed by `"Follower cannot join leader's match, releasing to independent matchmaking"`.

## Fix
Make the poll **patient**: treat "leader between matches" as *keep waiting*, and poll up to a bounded budget (**default 60s**) for the leader to land in a joinable lobby before falling back to solo. 60s comfortably covers a slow ("potato") leader.

- `presence == nil` now `continue`s instead of returning.
- Added an overall `pollDeadline` so the loop still terminates (it previously relied on that bail for one of its exits).
- Poll interval / max duration are per-pipeline fields with zero-value fallback to `pollFollowIntervalDefault` (3s) / `pollFollowMaxDurationDefault` (60s) — so it's unit-testable without multi-second waits, with **no global state**.
- Context cancellation still ends the poll promptly (covered by a test).

## Tests (red→green)
- `TestPollFollowPartyLeader_WaitsForLeaderBetweenMatches` — a permanently between-matches leader; asserts the poll spends its (tiny, overridden) budget waiting instead of bailing on the first tick. **Without the fix it takes the hardcoded 3s and fails the assertion** (verified: `3.0s is not less than 1s`).
- `TestPollFollowPartyLeader_StopsAtContextCancel` — a canceled context ends the poll promptly despite the long budget.

## Companion
Pairs with #508 (self-follow guard). Both touch the follow path but in different regions; the self-follow ghost case is handled there, the genuine-slow-leader case here.

### DIFF
--- a/server/evr_lobby_find.go
+++ b/server/evr_lobby_find.go
@@ -1717,22 +1717,37 @@ func (p *EvrPipeline) pollFollowPartyLeader(ctx context.Context, logger *zap.Log
 	}
 	pollDeadline := time.Now().Add(maxPollDuration)
 
-	for {
+	// pollWait sleeps for one poll interval, bounded by the time remaining in
+	// the overall budget so the total wait can never overshoot maxPollDuration
+	// by a full interval. It reports done=true when the caller should stop
+	// polling — because the budget is exhausted or the context was canceled —
+	// and converged=true when the matchmaker placed the follower during a
+	// canceled wait.
+	pollWait := func() (converged, done bool) {
+		remaining := time.Until(pollDeadline)
+		if remaining <= 0 {
+			logger.Debug("Poll budget exhausted; leader did not become joinable, releasing to independent matchmaking")
+			return false, true
+		}
+		wait := pollInterval
+		if wait > remaining {
+			wait = remaining
+		}
 		select {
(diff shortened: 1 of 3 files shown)


thesprockee:
Addressed the review: both waits could overshoot `maxPollDuration` because each slept a full `pollInterval` before checking the deadline. Extracted a `pollWait` closure used by the poll wait **and** the settle wait that bounds each sleep by `time.Until(pollDeadline)` and stops immediately when the budget is exhausted or the context is canceled — so the total wait can no longer exceed `maxPollDuration` regardless of how the interval/budget are configured.

PR #510, opened 2026-07-04
fix(party): make JoinPartyGroup transactional to kill ghost members

thesprockee:
## The ghost

Party **ghost members** originate in `JoinPartyGroup` (`server/evr_lobby_group.go`), which was **not transactional**. For an already-existing open group-name party it calls `ph.JoinRequest(&presence)`, which for an OPEN party **synchronously inserts the session into `ph.members`** (`party_handler.go:121-129`). It THEN calls `tracker.Track(...)`.

When that `Track` **fails** — which it does when the session's context is already canceled (a dropped/dying connection; `LocalTracker.Track` guards on `ctx.Err()`) — the function sends an error envelope and returns an error, but **never removes the member it just added** and **never deletes a party it just created**.

That orphaned member has membership but **no party-stream presence**, so the disconnect-eviction path (which is entirely stream-driven: session close → tracker untrack → `partyLeaveListener` → `partyRegistry.Leave` → `ph.Leave`) can never fire for it. It becomes a permanent ghost — later promoted to leader via `Oldest()` and leading a party for hours.

Confirmed against a real production ghost: session `bb06b40c`, ~350ms alive, then **2.7h as an unevictable party leader**.

Worse, the two callers in the still-running social-find loop (`server/evr_lobby_find.go`) **swallowed the error** (`if lobbyGroup, _, err := JoinPartyGroup(...); err == nil && ...`), so the failure was silent.

## The fix — transactional + error-visible

**`JoinPartyGroup`:**
- **Fail fast:** if `session.Context().Err() != nil` (session already dying), return an error immediately, before creating a party or adding any member. This is the common case for the ghost — a re-entrant call on a dead session.
- **Roll back on Track failure:** if `Track` returns `!success`, undo any state this call created before returning:
  - member added via `ph.JoinRequest` → `ph.members.Leave([]*Presence{&presence})`
  - party created via `GetOrCreateByGroupName` → `partyRegistry.Delete(ph.ID)`

The restored invariant: **`ph.members` must never retain a session that is not tracked on the party stream.** The just-added member is guaranteed not to be the leader at rollback time (leader assignment only happens via the tracker Join event, which requires a successful Track), so `ph.members.Leave` is the precise inverse of `JoinRequest` — no leader/stop reshuffle needed. The healthy success paths are byte-identical.

**Error visibility (`evr_lobby_find.go`, both call sites):** stop swallowing the error — log at `warn` with username / sid / group name / error, per the log-discipline rule (`warn` = someone should look). Control flow is otherwise unchanged; a failed join still falls through as before.

## Tests (TDD, red → green)

New `server/evr_lobby_group_ghost_test.go`. A `ghostCtxTracker` mirrors the real `LocalTracker` guard (Track fails when the session ctx is canceled); a `ghostFailTracker` fails Track on a live session to exercise the rollback deterministically.

| Test | Asserts | Pre-fix | Post-fix |
|---|---|---|---|
| `NoGhost_OnDeadSessionJoin` | dead-session join → error, no ghost in `ph.members` | FAIL (member lingers) | PASS |
| `NoOrphanParty_OnDeadSessionCreate` | dead-session create → error, no orphan party in registry | FAIL (orphan party) | PASS |
| `Rollback_RemovesMember_OnTrackFailure` | live session, Track fails → member rolled back | FAIL (ghost) | PASS |
| `Rollback_DeletesParty_OnTrackFailure` | live session, Track fails on create → party deleted | FAIL (orphan) | PASS |
| `HappyPath_JoinsExistingParty` | healthy join succeeds, member present | PASS | PASS |

Regression: full `-run 'Follow|Party|JoinPartyGroup|TryFollow|LobbyGroup'` party/follow suite passes (151 tests). The only failures are pre-existing/environmental and unrelated to this change: `TestGroupEntriesPartyAtomicity/party_exactly_fills_candidate` fails identically on the untouched base HEAD `8979ec6d5`, and a full-runtime InitModule test aborts on the unset `DISCORD_BOT_TOKEN`.

Gates: `gofmt`, `GOWORK=off go vet ./server/...`, `GOWORK=off go build ./...`, targeted `-race` tests — all clean.

### DIFF
--- a/server/evr_lobby_find.go
+++ b/server/evr_lobby_find.go
@@ -778,7 +778,15 @@ func (p *EvrPipeline) lobbyFindOrCreateSocial(ctx context.Context, logger *zap.L
 		// tracker-read priority join path.
 		if ws, ok := session.(*sessionWS); ok {
 			if lobbyParams.PartyGroupName != "" && lobbyParams.PartyGroupName != "tablet" {
-				if lobbyGroup, _, err := JoinPartyGroup(ws, lobbyParams.PartyGroupName, lobbyParams.CurrentMatchID); err == nil && lobbyGroup != nil {
+				lobbyGroup, _, err := JoinPartyGroup(ws, lobbyParams.PartyGroupName, lobbyParams.CurrentMatchID)
+				if err != nil {
+					logger.Warn("Failed to join party group in social lobby path",
+						zap.String("username", ws.Username()),
+						zap.String("sid", ws.ID().String()),
+						zap.String("group_name", lobbyParams.PartyGroupName),
+						zap.Error(err))
+				}
+				if err == nil && lobbyGroup != nil {
 					if reservationMatchID, found := p.findReservation(ctx, logger, ws, lobbyGroup); found {
 						logger.Info("Found party reservation in social lobby path, joining directly",
 							zap.String("reservation_mid", reservationMatchID.String()))
@@ -806,6 +814,13 @@ func (p *EvrPipeline) lobbyFindOrCreateSocial(ctx context.Context, logger *zap.L
 			ws, ok := session.(*sessionWS)
 			if ok {
 				lobbyGroup, _, err := JoinPartyGroup(ws, lobbyParams.PartyGroupName, lobbyParams.CurrentMatchID)
+				if err != nil {
+					logger.Warn("Failed to join party group in priority join path",
+						zap.String("username", ws.Username()),
+						zap.String("sid", ws.ID().String()),
+						zap.String("group_name", lobbyParams.PartyGroupName),
+						zap.Error(err))
+				}
 				if err == nil && lobbyGroup != nil {
 					leader := lobbyGroup.GetLeader()
 					if leader != nil && leader.SessionId != session.ID().String() {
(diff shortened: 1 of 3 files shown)


thesprockee:
Addressed both Copilot comments (commit `461c8270e`):
1. **Delete race** — the Track-failure rollback now guards the party delete on `ph.members.Size() == 0`, so a party another session legitimately joined is never deleted out from under them. The creator path skips `JoinRequest`, so on Track failure the creator isn't in `ph.members` and `Size()` reflects only committed co-members. Race-safety: `PartyPresenceList` Size/Join share the list's mutex, so a committed member is read consistently — the reported eviction is fully closed. A registry lock wouldn't help (the registry uses sharded maps; joins serialize on the PartyHandler, not the registry); full atomicity would need `ph.stopped`-flag coordination, a broader change I did not take. The residual (a joiner between GetOrCreate and JoinRequest exactly overlapping this rollback) is strictly narrower than the bug fixed and can never delete a party under a committed member — documented in the code comment. New test `TestJoinPartyGroup_Rollback_PreservesParty_WithCoMember_OnTrackFailure` (red→green); empty-party deletion still tested.
2. **Misleading message** — `"Error tracking party creation"` → `"Error tracking party membership"` (fires on both create and join paths).

PR #511, opened 2026-07-04
fix(party): unify group-name parties onto slot-reservation machinery

thesprockee:
## The dead signal path

Slot reservations hold seats so a party lands together in a social lobby. Two mechanisms exist; only one worked for real clients:

- **Placeholder path (works):** the leader appends placeholder reservations at its own social-lobby join.
- **Signal path (was dead in prod):** `createPartyReservations` -> `SignalCreatePartyReservations`, triggered in `lobbyEntrantConnected` **only when `params.currentPartyID != uuid.Nil`**.

`currentPartyID` was set **exclusively** by the native/SNS party handlers (`snsPartyTrackAndJoin`). Real clients never form native/SNS parties (0 SNS party messages in prod) — they form **group-name parties** via `LobbyGroupName` -> `JoinPartyGroup`, which never set `currentPartyID`. So the signal reservation path never fired for the only party type in use.

## The fix — two entry points, no new mechanism

The correct, party-type-independent rule: *a player in a party whose leader is currently in a social lobby should have a seat reserved for them in that lobby.*

1. **Populate `currentPartyID` in `JoinPartyGroup`** (mirroring `snsPartyTrackAndJoin`). This alone makes the existing leader-connect trigger (`lobbyEntrantConnected`) fire for group parties, and enables the matchmaker cancel path for them. Field comment at `evr_session_parameters.go` updated (no longer "SNS party" only).
2. **On-join reservation trigger:** a member that joins the party AFTER the leader is already in a social lobby is missed by the one-shot leader-connect trigger. `JoinPartyGroup` now calls the existing single-member helper `createReservationForNewPartyMember` for the non-leader joiner. That helper already gates on "leader is in a social match" (reusing the `StreamModeService` -> match -> `MatchLabelByID` -> `IsSocial()` lookup), so no parallel mechanism was introduced.

The placeholder path is left intact. The change is localized to the success path of `JoinPartyGroup` (kept clear of PR #510's ctx-fail-fast / rollback region for a trivial merge).

## Consumers verified

- `snsPartyLeaveCleanup`: only reachable from the SNS message handlers, which group clients never trigger — dormant, unaffected.
- Matchmaker cancel (`lobbyPendingSessionCancel`): now correctly cancels the whole party when a group member cancels (group parties previously missed this — a correctness improvement, per spec).

## Tests (TDD, red -> green)

New `server/evr_lobby_group_reservation_test.go`:

- `TestJoinPartyGroup_PopulatesCurrentPartyID` — red: `currentPartyID` stayed `uuid.Nil`; green: equals the registry party ID.
- `TestJoinPartyGroup_LeaderInSocial_CreatesReservationForJoiner` — red: "Condition never satisfied" (no reservation signal); green: a `SignalCreatePartyReservations` fires for the joining member.
- `TestJoinPartyGroup_LeaderInArena_NoReservation` and `TestJoinPartyGroup_LeaderNotInLobby_NoReservation` — no spurious reservation (green before and after).

Regression (`GOWORK=off go test ./server/ -race -run 'Follow|Party|Reservation|JoinPartyGroup|Matchmak|LobbyGroup'`): the failing set is **identical** to the pristine `main` baseline (environmental: file-fixture `Characterization*` matchmaker tests, `LobbyCapacity_*`, `GroupEntriesPartyAtomicity` — all fail without this change too). Zero new failures.

Gates: `gofmt` clean, `GOWORK=off go vet ./server/` clean, `GOWORK=off go build ./...` clean, targeted `-race` tests green.

### DIFF
--- a/server/evr_lobby_group.go
+++ b/server/evr_lobby_group.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	"errors"
 
 	"github.com/gofrs/uuid/v5"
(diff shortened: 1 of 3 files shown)


thesprockee:
Superseded by #513, which completes this work: reconciled onto #510 (transactional JoinPartyGroup — the `currentPartyID` set is now placed below the rollback so a failed join never sets it) + #512 (reservation dedup), adds the D3 clear-on-leave this draft was missing, and carries the full BAC test manifest. This draft's approach and test harness were reused in #513. Closing in favor of #513.

PR #512, opened 2026-07-05
fix(match): make CREATE-side reservation dedup UserID-keyed

thesprockee:
## Root cause: sessionID-keyed create, UserID-stable identity

Match slot reservations live in `state.reservationMap` keyed by **sessionID**. But an EVR client gets a **new nakama sessionID on every WebSocket reconnect** (`server/socket_ws.go:126` — `sessionID := uuid.Must(sessionIdGen.NewV1())`), while the **UserID** from the auth token is stable.

The **consume** side already handles this: `MatchJoinAttempt` does `LoadAndDeleteReservationRaw(sessionID)` then falls back to `LoadAndDeleteReservationByUserIDRaw(userID)` (`evr_match.go:416-421`), documented in `evr_match_label.go:97-98`. The **create** side had **no UserID guard**.

## The reconnect seat-leak trace

1. Member M reserved under `sidA`.
2. M's client reconnects → new session `sidB`; a create path re-fires for M under `sidB`.
3. Create-dedup checks `reservationMap[sidB]` → MISS (orphan is under `sidA`) → creates a **second** entry.
4. Both count toward capacity: `OpenSlots() = MaxSize - Size - ReservationCount`, and `ReservationCount` counts every non-expired `reservationMap` entry (`rebuildCache`, `evr_match_label.go:407-432`).
5. On M's real join under `sidB`, `LoadAndDeleteReservationRaw(sidB)` consumes `sidB` and returns early — the UserID fallback is never reached — so the `sidA` **orphan is never reclaimed** and holds a phantom seat until its 5-min expiry.

Net effect: a spurious `LobbyFull`/`ReservationViolated` rejection for an unrelated joiner during the window.

## The fix: one shared UserID-keyed upsert helper

New `MatchLabel.upsertReservationByUserID(reservation)` (`evr_match_label.go`) removes any existing `reservationMap` entry for the same UserID (any sessionID), then inserts the new entry under its sessionID — making CREATE symmetric with the UserID-aware consume. Deduped on `EvrMatchPresence.UserID` (the auth-token UUID, stable across reconnects — the exact field the consume path already uses). A nil UserID falls back to sessionID-only keying so identity-less members are never collapsed.

Wired into both create sites:
- **`SignalCreatePartyReservations`** (`evr_match.go`) — now skips members already present **by UserID** and upserts by UserID.
- **`MatchJoinAttempt` placeholder write-in** (`evr_match.go`) — now upserts by UserID (previously a raw sessionID insert with no dedup).
- **Non-leader `MatchLeave` clear** (`evr_match.go`) — now deletes the leaving member's reservation **by UserID**, closing the same orphan blind spot.

## Blast-radius verified safe

- `restoreSlotReservation` and the `MatchInit`/`SignalPrepareSession` seed are **untouched** (raw sessionID inserts). The upsert's remove-then-insert happens in one call, so a same-member restore/re-seed can never net to zero — regression-tested.
- The separate, already-UserID-keyed `reconnectReservations` map is **not touched**.
- Guild isolation untouched: dedup is per-match `reservationMap`; no `GroupID` normalization.
- Both feeder paths (`createPartyReservations`, `createReservationForNewPartyMember`) supply a real auth UserID (`pp.UserID` / `memberSession.userID`).

## Red → green evidence

New `server/evr_reservation_dedup_test.go` drives the real `SignalCreatePartyReservations` handler via `MatchSignal`:

**Against base (RED):**
```
--- FAIL: TestReservationDedup_ReconnectDoesNotDoubleCount
    expected exactly 1 reservation after reconnect, got 2 (phantom seat leak)
    stale reservation under old session ID sidA is still present (orphan not reclaimed)
    expected ReservationCount=1 (one held seat), got 2
--- FAIL: TestReservationDedup_OpenSlotsCorrectAfterReconnect
    expected OpenSlots()=11 (one held seat), got 10 (phantom seat still counted)
```

**After fix (GREEN):**
```
--- PASS: TestReservationDedup_ReconnectDoesNotDoubleCount
--- PASS: TestReservationDedup_OpenSlotsCorrectAfterReconnect
--- PASS: TestReservationDedup_SameMemberReseedKeepsSeat      (restore keeps exactly one seat)
--- PASS: TestReservationDedup_DistinctMembersUnaffected      (two users not collapsed)
```

Checks run: `gofmt -l` clean, `GOWORK=off go vet ./server/...` clean, `GOWORK=off go build ./...` clean, tests with `-race`. The targeted suite (`-run 'Reservation|Capacity|MatchJoin|OpenSlots|Party'`) shows the **identical** pre-existing failure set on this branch and on untouched base (`TestLobbyCapacity_*`, `TestEvrMatch_MatchJoinAttempt_Counts`, `TestGroupEntriesPartyAtomicity/party_exactly_fills_candidate`) — **no new failures** introduced.

### DIFF
--- a/server/evr_match.go
+++ b/server/evr_match.go
@@ -533,7 +533,10 @@ func (m *EvrMatch) MatchJoinAttempt(ctx context.Context, logger runtime.Logger, 
 			// Reserve spots for party members for 5 minutes
 			reservation.Expiry = time.Now().Add(time.Minute * 5)
 		}
-		state.reservationMap[sessionID] = reservation
+		// Upsert by user ID so a member who reconnected (new session ID) does not
+		// leave a phantom reservation under their stale session ID. Symmetric with
+		// the UserID-aware consume path (LoadAndDeleteReservationByUserIDRaw).
+		state.upsertReservationByUserID(reservation)
 		state.joinTimestamps[sessionID] = time.Now()
 	}
 
@@ -880,9 +883,15 @@ func (m *EvrMatch) MatchLeave(ctx context.Context, logger runtime.Logger, db *sq
 					}
 				} else {
 					// Non-leader is leaving: clear only THEIR OWN reservation.
-					if _, exists := state.reservationMap[mp.GetSessionId()]; exists {
-						delete(state.reservationMap, mp.GetSessionId())
-						cleared = 1
+					// Match by user ID, not just session ID: a member who
+					// reconnected holds their reservation under a NEW session ID
+					// (see upsertReservationByUserID), so a session-ID-only delete
+					// would orphan the old entry.
+					for sid, r := range state.reservationMap {
+						if r.Presence.UserID == mp.UserID {
+							delete(state.reservationMap, sid)
+							cleared++
+						}
 					}
 				}
 				if cleared > 0 {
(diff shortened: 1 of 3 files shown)


thesprockee:
Addressed both Copilot comments (commit `e16d105cf`) — nil-UserID consistency with `upsertReservationByUserID`:
1. **MatchLeave non-leader clear** — when `mp.UserID == uuid.Nil`, now deletes only the leaver's own reservation by session ID instead of wiping every nil-UserID reservation.
2. **`SignalCreatePartyReservations` alreadyPresent** — when `member.UserID == uuid.Nil`, compares by session ID instead of UserID, so distinct nil-UserID members no longer collapse and a needed reservation isn't skipped.
Non-nil (reconnect-orphan) behavior untouched. New tests `TestReservationDedup_NilUserIDLeaveOnlyClearsOwn` and `_NilUserIDCreateNotSkipped` (red→green); existing dedup tests stay green.

PR #513, opened 2026-07-05
fix(party): unify group-name parties onto slot-reservation machinery

thesprockee:
Completes the party-reservation unification. **Stacks on #510 and #512** — merge those first (this branch was built and tested on a base that includes both).

Supersedes #511 (an earlier draft of this same work that only had 2 of the 3 changes and was built pre-#510/#512).

## Problem
Real clients form parties via `LobbyGroupName` → `JoinPartyGroup`, which never set `params.currentPartyID`. The signal-based slot-reservation path (`lobbyEntrantConnected` → `createPartyReservations`) is gated on `currentPartyID != nil`, set **only** by the native/SNS party handlers — which real clients never use (0 SNS party messages in prod). So the reservation path was dead for the only party type in use, and the matchmaker cancel-all path treated group members as solo.

## Fix (D-numbers from the validated design)
- **D2 — `currentPartyID` for group parties** (`evr_lobby_group.go`): set `params.currentPartyID = ph.ID` in `JoinPartyGroup`, placed **after #510's Track-failure rollback** so a rolled-back join never sets it. `currentSNSPartyID` stays 0 (every reader is SNS-gated). This lights up the reservation trigger **and** the matchmaker cancel-all for group parties.
- **E2 — on-join reservation** (`evr_lobby_find.go` `configureParty`): a member who joins after the leader is already in a social lobby gets a seat via the existing `createReservationForNewPartyMember` (self-guards leader-in-social; skips arena/combat), dispatched `go …WithoutCancel`.
- **D3 — clear on leave** (`evr_lobby_session.go`): factored `clearPartyParams` out of `snsPartyLeaveCleanup` and call it at the spectator group-leave so an ex-member doesn't carry stale party state. Disconnect needs no explicit clear (params die with the session — pinned by a test).
- Log-discipline: `createReservationForNewPartyMember`'s per-join success log downgraded `Info` → `Debug`.

## Tests (all red→green, `-race`)
BAC-1 (currentPartyID set + rolled-back join stays nil), BAC-2 (E1 fires for group), BAC-3 (E2 on-join), BAC-4 (spectator leave clears), BAC-4b (disconnect self-cleaning), BAC-5 (E1+E2+reconnect → one reservation, inherited from #512), BAC-6 (cancel-all). Guardrails green: #510 ghost tests, #512 dedup tests, guild isolation untouched.

Design + plan: developed through a design → validate → verify → plan → implement → verify pipeline.

### DIFF
--- a/server/evr_lobby_find.go
+++ b/server/evr_lobby_find.go
@@ -396,6 +396,20 @@ func (p *EvrPipeline) configureParty(ctx context.Context, logger *zap.Logger, se
 		zap.String("leader_uid", leader.GetUserId()),
 		zap.String("current_match_id", lobbyParams.CurrentMatchID.String()))
 
+	// E2 (D1): if this session is NOT the leader, a member may have joined a
+	// party whose leader is ALREADY settled in a social lobby (I3's timing case),
+	// which the one-shot leader-connect trigger (E1) cannot cover. Reserve a seat
+	// for them now. createReservationForNewPartyMember self-guards leader-is-self
+	// (:757), leader-in-a-match (:768) and leader-in-social (:778) internally, so
+	// it is a no-op unless the leader is currently in a social match — do NOT
+	// duplicate those checks here. Dispatched WithoutCancel so it survives this
+	// session's find-loop teardown, matching the SNS caller (snsPartyJoinRequest).
+	// Idempotent via #512's UserID-keyed dedup, so a repeat fire yields one
+	// reservation (BAC-5).
+	if !isLeader {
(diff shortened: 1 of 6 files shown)


thesprockee:
The two nil-UserID comments on this PR are in code inherited from #512; fixed there (commit `e16d105cf`) so the fix lives with the `upsertReservationByUserID` helper it's consistent with. This branch has been rebuilt on the updated #510 + #512 (head `3b775ac86`) so its diff carries the corrected code. Stacks on #510 + #512 — merge those first.

PR #564, opened 2026-08-08
feat(detect): optional first-sight login reject on a machine-fingerprint match

thesprockee:
`Refs #516` — its **top-ranked** fix direction: *"First-sight reject/flag on exact full `system_profile` match to a currently-disabled account."*

## What was actually missing

Detection already worked. After #551 a specific system profile is a strong signal, so `filterStrongAlts` does not suppress it and the delayed kick fires. **The gap was only timing** — the current path admits the session and disables + kicks it after a jittered 1–4 minutes.

So this adds the first-sight path and **leaves the delayed one intact behind it**.

## Off by default — a trade, not an upgrade

The delayed kick costs a few minutes of presence and buys **ambiguity about which signal caught them**. Rejecting at login is immediate but tells an evader exactly what to change, on the login that carried it — which is the feedback loop that makes rotation cheap.

That is an operator's call, so the operator makes it. `RejectDisabledAlternatesOnMachineMatch`, default false. Same posture as `MinimumNakamaAccountAgeDays` in #562.

## Narrower than `filterStrongAlts`, deliberately

`filterStrongAlts` keeps an alt if **any** match item is strong — so an alt linked only by a residential IP on a non-CGNAT range survives it. This asks the single question item 1 is about: *is this the same machine?*

Only a full system profile answers that. An IP is a household and a reboot rotates it; an HMD serial is rotated by editing a config. Neither triggers this gate.

## The bucket guards are the load-bearing part

**This rejects logins.** A bucket here does not mis-tag someone — it locks out real players en masse. Two exclusions, both proven rather than assumed:

| excluded | why |
|---|---|
| `Unknown::::::::0::0::0::0` | Every account that logs in without `SystemInfo` emits this identical string. **The bucket that nearly shipped with #551.** |
| `Meta Quest 3::…` and siblings | Stock hardware describes a large share of the player base. |

**Mutation-proven.** Removing the degenerate guard:

```
--- FAIL: TestIsMachineFingerprint/DegenerateProfileIsNotAFingerprint
--- FAIL: TestIsMachineFingerprint_NilDetectorNarrows
```

Removing the commodity guard:

```
--- FAIL: TestIsMachineFingerprint/CommodityHeadsetIsNotAFingerprint
--- FAIL: TestIsMachineFingerprint_NilDetectorNarrows
--- FAIL: TestMachineMatchedAlts
```

## An unavailable classifier must narrow, never widen

`TestIsMachineFingerprint_NilDetectorNarrows` pins this directly. If the CGNAT detector is unavailable, the commodity check cannot run — and the wrong reading of that ("nothing is commodity") would turn **every Quest profile into a machine match** the moment settings failed to load, on a gate that rejects logins.

So a nil detector only skips the commodity check; it never makes `isMachineFingerprint` return true on its own. And the degenerate guard does not consult settings at all, so it holds even then.

## Failure posture

- **Fails open on a lookup error**, and logs it. This gate is an *addition* to the delayed kick, not a replacement — a failed `AccountsGetId` must not lock out a legitimate player when the existing path still runs behind it.
- `ignoreDisabledAlternates` is honoured exactly as the delayed path honours it, so an account a moderator has cleared stays cleared.
- Evaluated **before** `filterStrongAlts` narrows `firstIDs`, so the two filters cannot mask one another.
- The rejection reuses `AccountDisabledError` — the same message a disabled account gets. Telling an evader which signal caught them is the cost this setting trades against; the detail goes to the log and the metric, where moderators can see it.

## Also

`disabledAccountIDs` extracted rather than inlined a fourth time. It returns an error instead of swallowing one, so each caller picks its own failure posture — the two that gate on it choose differently, and both are deliberate.

New metric: `login_rejected_machine_matched_disabled_alt`.

## Still open on #516

Item 2 (subnet/ASN) is unchanged and still needs the bucket question answered against production data — see [this comment](https://github.com/EchoTools/nakama/issues/516#issuecomment-5225217000), where the datacenter data turned out to already exist in both providers.

And the operational question this PR does not answer: **is `KickPlayersWithDisabledAlternates` enabled in production?** If it is not, the delayed path this sits in front of has never fired either.

Full `just test-audit` green: 3387 pass, 130 skip, 0 fail.

Co-authored-by: Andrew Bates <a@sprock.io>

### DIFF
--- a/server/evr_authenticate_alts.go
+++ b/server/evr_authenticate_alts.go
@@ -4,10 +4,46 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"strings"
 
 	"github.com/heroiclabs/nakama-common/api"
 	"github.com/heroiclabs/nakama-common/runtime"
 )
+
+// disabledAccountIDs returns which of the given user IDs belong to accounts
+// that are currently disabled.
+//
+// Returns an error rather than swallowing one, so callers choose their own
+// failure posture. The two that gate on this choose differently and both are
+// deliberate: the login rejection fails open (a lookup failure must not lock a
+// legitimate player out, and the delayed kick still runs behind it), while a
+// caller that already admitted the session can afford to be stricter.
+func disabledAccountIDs(ctx context.Context, nk runtime.NakamaModule, userIDs []string) ([]string, error) {
+	if len(userIDs) == 0 {
+		return nil, nil
+	}
+
+	accounts, err := nk.AccountsGetId(ctx, userIDs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get accounts for user IDs %v: %w", userIDs, err)
+	}
+
+	disabled := make([]string, 0, len(accounts))
+	for _, a := range accounts {
+		// A zero DisableTime is "not disabled", not "disabled at the epoch" --
+		// both the nil and the zero form appear in practice.
+		if a.GetDisableTime() == nil || a.GetDisableTime().AsTime().IsZero() {
+			continue
+		}
+		if a.GetUser() != nil {
+			disabled = append(disabled, a.GetUser().GetId())
+		}
+	}
+	if len(disabled) == 0 {
+		return nil, nil
+	}
+	return disabled, nil
+}
 
 type AlternateSearchMatch struct {
 	OtherUserID string   `json:"other_user_id"`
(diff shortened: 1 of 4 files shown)


thesprockee:
## `RejectDisabledAlternatesOnMachineMatch` could never have fired — enabled or not

Verified against `main` @ `6e9e5dbb8` (current HEAD). This is not "the setting was off". The gate was mechanically unreachable, and would have stayed unreachable if someone had turned it on.

### The chain

`isMachineFingerprint` (`server/evr_authenticate_alts.go:237-248 @ 6e9e5dbb8`) returns false if **either** guard trips:

```go
if matchIgnoredAltPattern(item) {
	return false
}
if detector != nil && detector.IsWeakSignal(item) {
	return false
}
```

Both tripped, for every system profile:

- `IsWeakSignal` (`server/evr_cgnat.go:158, loop at :173-177 @ 6e9e5dbb8`) matched every non-IP string, because production's `cgnat.commodity_profile_prefixes` carried a `""` alongside the real `"Meta Quest N::"` entries and `strings.HasPrefix(x, "")` is always true. That is #589.
- `matchIgnoredAltPattern` (`server/evr_authenticate_history.go:50`, the `IsWeakSignal` branch at `:64 @ 6e9e5dbb8`) drops any item where `IsWeakSignal` is true **and** `net.ParseIP` fails and the string is neither `""` nor `"unknown"`. A full system profile satisfies all three, so it was dropped there too.

So `isMachineFingerprint` returned false at `:241` before it ever reached `:244`, `machineMatchedAlts` (`server/evr_authenticate_alts.go:257-277`) always returned nil, and the body of the gate at `server/evr_pipeline_login.go:657-658 @ 6e9e5dbb8` was never entered — `len(machineIDs) > 0` was never true.

### And a second, independent reason

Even with `isMachineFingerprint` repaired, there would have been nothing for it to read. `loginHistoryCompare` filters every candidate item through `matchIgnoredAltPattern` before emitting it (`server/evr_authenticate_alts.go:336 @ 6e9e5dbb8`), so `LoginHistory.AlternateMatches[...].Items` could not *contain* a system profile at all. Production measurement, 2026-09-08: **7,254 of 7,254** stored alt links rested on a shared IP; zero carried an XPID, HMD serial or system profile. See #589 for the full mechanism and evidence table.

### The correction to the record

[This comment, 2026-08-08](https://github.com/EchoTools/nakama/issues/516#issuecomment-5225147688) said of item 1:

> **The mechanism already exists and already fires.**

The first half held. The second did not, and had not since before the PR was written. The reasoning in that comment — "`IsWeakSignal` classifies a system profile as weak only when it matches `CommodityProfilePrefixes` — which **defaults** to `{"Meta Quest 2::", ...}`" — was correct about the *seeded defaults* and wrong about the *stored value*, which is where the `""` lived. The source could not defend itself against it and nothing logged it.

### It also means the first question was the wrong one

Both this PR's description and [the follow-up](https://github.com/EchoTools/nakama/issues/516#issuecomment-5225350466) closed on:

> **Is `KickPlayersWithDisabledAlternates` actually enabled in production?** Still unanswered, and it is the first thing to check.

`RejectDisabledAlternatesOnMachineMatch` does not consult that setting at all — it is a separate gate on a separate path (`server/evr_pipeline_login.go:657` vs. `server/evr_runtime_event_user_authenticate.go:109`). Answering "yes" would not have made this gate fire. And the delayed path that question is about narrows through `filterStrongAlts` (`server/evr_runtime_event_user_authenticate.go:86`), which the same `""` collapsed to "linked by a non-CGNAT IP". So the operator question was downstream of a defect that made both answers moot.

The right first question was: *has this mechanism ever produced a non-empty result in production?* Neither gate emits a metric or a log line on the empty case, so nothing could answer it from the outside — which is the reason it went ~a month unnoticed.

### What is not claimed here

**There is no evidence either way about whether the gate works.** This says only that it has never run. It needs re-measuring once **both** of these are true:

1. #589's guard has landed (and/or the `""` is removed from the stored `Global/settings` record — the code guard makes it harmless but the value is still wrong), and
2. **the alt links have been recomputed.** The 7,254 existing edges were built by the broken filter and contain IPs only. Fixing `IsWeakSignal` does not retroactively add system profiles to stored `AlternateMatches` — `machineMatchedAlts` reads that stored data. Until the histories are re-compared, this gate keeps returning nil for exactly the same reason, with the defect fixed.

Only after both is a measurement of this gate meaningful.

Refs #589, #516. (This PR is merged and cannot be reopened; recording here because this is where the mechanism is documented.)

PR #579, opened 2026-08-20
fix: runtime fixes from the pre-release housekeeping sweep

thesprockee:
Two runtime defect fixes recovered from the Aug-19 campaign worktrees, reviewed and ruled out of the evr.324 release by the owner (they land post-release).

**1. Spectate fail-open (evr_lobby_session.go)** — a spectate request in a non-arena/combat mode assigned its rejection to a function-scoped `err` that was never read on the spectator path, so the request returned nil success AFTER the party teardown had already run. Fix: validate before teardown, return the error, retain the party. Origin: worktree-agent-af1fc576f37b0edd7 @ 8286bfeb4.

**2. Serverscore window threading (evr_lobby_builder_serverscore.go)** — `calculateServerScore` hardcoded its min/max/threshold window and the exported wrapper silently discarded caller-passed rttMin/rttMax. Fix: thread the parameters through; defaults byte-identical to the old constants, so default-path scores are unchanged. Only callers passing explicit MinRTT/MaxRTT see a behavior change — previously ignored, now honored. Test fix recovered alongside (six of seven score cases asserted nothing). Origin: worktree-agent-a23fc8f3b8f3e7905 @ b160574d5 + 1638f4e84.

**Gates:** lint 286, below the prior 290 baseline — LINT_BASELINE lowered to 286 in this series (the ground these fixes took). go vet/gofmt clean. Note: the two gopls diagnostics in evr_discord_appbot.go:3329/3337 are pre-existing and out of this PR's scope.

Post-merge: the four worktree-agent-* branches can be deleted (SHAs preserved here).


thesprockee:
Superseded. The content of this branch is merged into `main` (see `4e39d955b` and `9b9890f5d`), which is green: `just verify` all 6 checks, 3564 pass / 0 fail. Branch deleted.

